PR #599, opened 2014-12-08
Improve autocompletion

Fatme:
* Move build, livesync-cloud and prop commands to `commands`
* Implement completionData for build, livesync-cloud and prop commands

Should be merged after this https://github.com/telerik/mobile-cli-lib/pull/112
Helper PR for this story http://teampulse.telerik.com/view#item/280276


lib/commands/prop/prop-print.ts
```diff
@@ -0,0 +1,32 @@
...
+class PrintProjectCommandParameter implements ICommandParameter {
+	constructor(private $project: Project.IProject) { }
+
+	mandatory = false;
+
+	validate(validationValue: string): IFuture<boolean> {
+		return (() => {
+			this.$project.ensureProject();
+
+			return validationValue;
```
rosen-vladimirov: this should return true instead of validationValue

rosen-vladimirov: Or even better - return true in case there is validationValue

rosen-vladimirov:
:+1:  after fixing my comment

PR #635, opened 2014-12-17
Implement platform-specific .abignore files

teobugslayer:
Implement platform-specific .abignore files

Refs http://teampulse.telerik.com/view#item/281060

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -24,7 +24,7 @@ export class Project implements Project.IProject {
 	private static JSON_PROJECT_FILE_NAME_REGEX = "[.]abproject";
 	private static CONFIGURATION_FILE_SEARCH_PATTERN: RegExp = new RegExp(".*.abproject$", "i");
 	private static VALID_CONFIGURATION_CHARACTERS_REGEX = "[-_A-Za-z0-9]";
-	private static CONFIGURATION_FROM_FILE_NAME_REGEX =  new RegExp("^[.](" + Project.VALID_CONFIGURATION_CHARACTERS_REGEX + "+?)" + Project.JSON_PROJECT_FILE_NAME_REGEX + "$", "i");
+	private static CONFIGURATION_FROM_FILE_NAME_REGEX = new RegExp("^[.](" + Project.VALID_CONFIGURATION_CHARACTERS_REGEX + "+?)" + Project.JSON_PROJECT_FILE_NAME_REGEX + "$", "i");
 
 	public configurationSpecificData: IDictionary<IDictionary<any>>;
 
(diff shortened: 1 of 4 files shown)


lib/project.ts
```diff
@@ -164,7 +164,11 @@ export class Project implements Project.IProject {
 				return !isExcluded && !isSubprojectDir;
 			});
 
-			var ignoreFilesRules = this.$pathFilteringService.getRulesFromFile(path.join(projectDir, Project.IGNORE_FILE));
+			var ignoreFilesRules: string[] = [];
+			_.each(this.ignoreFilesConfigurations, (configFile) => {
```
teobugslayer: I am appending to the output string. But I will rewrite it for clarity.

lib/project.ts
```diff
@@ -270,6 +274,20 @@ export class Project implements Project.IProject {
...
 	}
 
+	public getBuildConfiguration(): string {
+		return options.release || options.r ? "Release" : "Debug";
+	}
+
+	private get ignoreFilesConfigurations(): string[] {
+		var configurations: string[] = [ Project.IGNORE_FILE ];
+		if(options.release || options.r) {
+			configurations.push("." + Project.DEBUG_CONFIGURATION_NAME + Project.IGNORE_FILE);
```
teobugslayer: Doh, a copy-paste mistake. Thanks for spotting!

teobugslayer:
Addressed review comments, @ligaz

lib/project.ts
```diff
@@ -149,7 +149,7 @@ export class Project implements Project.IProject {
 	}
 
 	private static IGNORE_FILE = ".abignore";
-	private static INTERNAL_NONPROJECT_FILES = [".ab", Project.IGNORE_FILE, "**/*.ipa", "**/*.apk", "**/*.xap"];
+	private static INTERNAL_NONPROJECT_FILES = [".ab", Project.IGNORE_FILE, ".*" + Project.IGNORE_FILE, "**/*.ipa", "**/*.apk", "**/*.xap"];
```
rosen-vladimirov: Project.IGNORE_FILE, ".*" + Project.IGNORE_FILE

can we merge these in:
```
"*" + Project.IGNORE_FILE
```

lib/project.ts
```diff
@@ -164,7 +164,10 @@ export class Project implements Project.IProject {
 				return !isExcluded && !isSubprojectDir;
 			});
 
-			var ignoreFilesRules = this.$pathFilteringService.getRulesFromFile(path.join(projectDir, Project.IGNORE_FILE));
+			var ignoreFilesRules = <string[]>_(this.ignoreFilesConfigurations)
```
rosen-vladimirov: Is this cast required? Can we use:
```Typescript
var ignoreFilesRules: string[] = _(this.ignoreFilesConfigurations)
```
(looks like I really don't like <...> cast syntax )

rosen-vladimirov:
:+1:  two minor comments, which are just for consideration, not show stoppers

lib/project.ts
teobugslayer: It is required, because without it the source does not compile. I think that the issue is due to insufficiently sophisticated typescript definitions.

lib/project.ts
teobugslayer: What if the user has a file named "pencho.abproject", should we exclude it as well?

rosen-vladimirov: I'm okay with the discussion we had, so you can leave the code as it is :)

PR #637, opened 2014-12-18
Replace underscore with lodash (finally!)

teobugslayer:


rosen-vladimirov:
run ci

lib/options.ts
```diff
@@ -2,8 +2,6 @@
 "use strict";
 
 import helpers = require("./common/helpers");
-
-import _ = require("underscore");
```
rosen-vladimirov: we need this import (of course change underscore with lodash). You have added a global _ to point to lodash in lib/common/bootstrap, which is loaded after options.ts, so at this point there's no _ and trying to execute npm install fails. Also options.ts has reference to common/helpers.ts, where _ is used as well, so we should have _ defined there as well.

rosen-vladimirov:
:+1:

PR #638, opened 2014-12-18
Allow emulate android command to wait forever using --timeout 0

teobugslayer:
Implements http://teampulse.telerik.com/view#item/280977


rosen-vladimirov:
:+1:

PR #639, opened 2014-12-18
Support for WP8 Ion

teobugslayer:
http://teampulse.telerik.com/view#item/262237


vendor/qrcode.js
```diff
@@ -1785,11 +1786,19 @@ var qrcode = function() {
...
+    dataUri += base64;
+    return dataUri;
+  };
+
+
+  var createImgTag = function(width, height, getPixel, alt) {
     var img = '';
     img += '<img';
     img += '\u0020src="';
     img += 'data:image/gif;base64,';
```
rosen-vladimirov: Do we need this line? Isn't it duplicated as in CreateDataUri we have it as well?

rosen-vladimirov:
:+1:  but I think you have to rebase :P

PR #643, opened 2015-01-05
Update package.json

rosen-vladimirov:
Update package.json to use v1.0.2.0 of icenium fibers fork and updated node requirements to exclude version 0.10.34.

Fixes http://teampulse.telerik.com/view#item/282812 and http://teampulse.telerik.com/view#item/282652


Fatme:
:+1:

PR #644, opened 2015-01-05
Fix failing android build

Fatme:
* Fixes "#<Object> has no method 'createDataUri'" error when appbuilder build command is executed -  createDataUri function was removed from the new version of qrcode-generator library and we need to extract dataUri from imgTag
* Remove qrcode.js  from vendor and use qrcode-generator from npm instead
* http://teampulse.telerik.com/view#item/282648


rosen-vladimirov:
:+1:

PR #647, opened 2015-01-07
Fix simulate command

Fatme:
Fixes http://teampulse.telerik.com/view#item/283107


rosen-vladimirov:
:+1:  after green build

PR #649, opened 2015-01-07
Unlock avd after app installation

rosen-vladimirov:
Update common lib where the following change is applied:
Add unlock of avd image after installation is completed for cases when it takes a lot of time. This way the application will be loaded immediately. Without this code, the user has to unlock the device manually and after that the launch will start.

http://teampulse.telerik.com/view#item/280997


teobugslayer:
:+1:

PR #654, opened 2015-01-08
Remove default Cordova plugins & variables

teobugslayer:
They should come from the templates instead.

Fixes http://teampulse.telerik.com/view#item/283210


Fatme:
:+1:

PR #657, opened 2015-01-09
Android Emulator now checks whether the SDK is correctly configured

teobugslayer:
Also gives a proper message to the user

Implements http://teampulse.telerik.com/view#item/279833


rosen-vladimirov:
:+1:

PR #658, opened 2015-01-09
Store user data & login cookie per-environment

teobugslayer:
See http://teampulse.telerik.com/view#item/283384

### DIFF
--- a/lib/login.ts
+++ b/lib/login.ts
@@ -16,6 +16,7 @@ export class UserDataStore implements IUserDataStore {
 	private user: any;
 
 	constructor(private $fs: IFileSystem,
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
:+1:  but this will require each customer to log on once as we'll change the cookie

teobugslayer:
@rosen-vladimirov - yes, this is described in the TP item.

teobugslayer:
This change will be visible to the users. Once they update appbuilder, the old files won't be found and they'll need to log in once. @ikoevska, should we document this?

rosen-vladimirov:
ah, sorry. 
Great work btw :) We'll all love this feature!!!

PR #659, opened 2015-01-12
Don't fail if the request to "http://freegeoip.net/json/" is not successful

Fatme:
Fixes http://teampulse.telerik.com/view#item/283398


rosen-vladimirov:
:+1:

tytodorov:
:+1:

PR #662, opened 2015-01-12
Print the correct server error message when 402 is returned

teobugslayer:
http://teampulse.telerik.com/view#item/283456

### DIFF
--- a/lib/service-util.ts
+++ b/lib/service-util.ts
@@ -66,6 +66,8 @@ export class ServiceProxy implements Server.IServiceProxy {
 			} catch(err) {
 				if (err.response && err.response.statusCode === 401) {
 					this.$userDataStore.clearLoginData().wait();
+				} else if (err.response && err.response.statusCode === 402) {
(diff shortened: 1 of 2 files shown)


lib/service-util.ts
```diff
@@ -66,6 +66,8 @@ export class ServiceProxy implements Server.IServiceProxy {
 			} catch(err) {
 				if (err.response && err.response.statusCode === 401) {
 					this.$userDataStore.clearLoginData().wait();
+				} else if (err.response && err.response.statusCode === 402) {
```
rosen-vladimirov: Maybe we can use switch in case it is expected to handle other codes in the future:
```Typescript
if (err.respone) {
    switch(err.response.statusCode) 
 ....
}

rosen-vladimirov:
:+1:  Just one minor suggestion for future improvements

lib/service-util.ts
teobugslayer: A good catch, but I will restruct it the third time it is needed.

rosen-vladimirov: Absolutely agree with this  :100:

PR #666, opened 2015-01-13
Usa shared appid option

Fatme:


rosen-vladimirov:
:+1:  after green build

PR #672, opened 2015-01-15
Minor code refactor for the plugins-service

Mitko-Kerezov:
getInstalledPlugins and getInstalledPluginsEnabledAtLeastInOneConfiguration methods were almost identical

### DIFF
--- a/lib/services/plugins-service.ts
+++ b/lib/services/plugins-service.ts
@@ -30,30 +30,12 @@ export class PluginsService implements IPluginsService {
 	}
 
 	public getInstalledPlugins(): IPlugin[] {
-		var corePlugins: any = null;
-		if(options.debug || options.d) {
-			corePlugins = this.$project.getProperty(PluginsService.CORE_PLUGINS_PROPERTY_NAME, "debug");
-		} else if(options.release || options.r) {
-			corePlugins = this.$project.getProperty(PluginsService.CORE_PLUGINS_PROPERTY_NAME, "release");
-		} else {
-			corePlugins = _.intersection(this.$project.getProperty(PluginsService.CORE_PLUGINS_PROPERTY_NAME, "debug"), this.$project.getProperty(PluginsService.CORE_PLUGINS_PROPERTY_NAME, "release"));
-		}
-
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
Why do you need getInstalledPlugins and getInstalledPluginsEnabledAtLeastInOneConfiguration at all? You can just replace the place where they are called with:
```Typescript
this.getAllInstalledPlugins(true); // or false parameter
```

Also passing boolean parameter to a method, does not show you the real meaning of this parameter. So you have to go to method's implementation in order to understand what kind of value you should pass when you want to use the method. In such cases we prefer the following:
```Typescript
getAllInstalledPlugins(enabledInConfig: {isEnabledAtLeastOnce: boolean}) {
     //....
     if(enabledInConfig.isEnabledAtLeastOnce) { 
     } else { 
     }
}

// when calling the method you should use:
this.getAllInstalledPlugins({isEnabledAtLeastOnce: true}); // or false of course
```
This way you can easily call the method and understand what parameter you should pass without checking its implementation.

Note: I've used sample variable names, you can use names you prefer :)

teobugslayer:
Or you can simply* use an enum.

Caveat Emptor: "simply" is rarely applciable to typescript enums.

Mitko-Kerezov:
Addressed the comments and trying out entirely different approach
Ping @teobugslayer @rosen-vladimirov

lib/services/plugins-service.ts
```diff
@@ -257,5 +239,18 @@ export class PluginsService implements IPluginsService {
 
 		return plugin;
 	}
+
+	private getAllInstalledPlugins(configuration: {getPluginsWithoutOptions: (...args: any[]) => any[]}): IPlugin[] {
```
teobugslayer: The parameter name is inexact. What about something similar to "operation"?

rosen-vladimirov: yes, I was confused of the name as well, operation seems better for me :) 
btw great idea to pass the operation instead of boolean value

teobugslayer:
:+1:

PR #673, opened 2015-01-16
Check for future projectVersion and refuse working

teobugslayer:


rosen-vladimirov:
:+1:  after fixing the unit test

PR #676, opened 2015-01-16
Share the dev-postinstall command between the CLIs

teobugslayer:
refs NativeScript/nativescript-cli#171


rosen-vladimirov:
:+1:

PR #678, opened 2015-01-16
The help says that one can publish NativeScript projects to AppManager

teobugslayer:
See http://teampulse.telerik.com/view#item/283182


rosen-vladimirov:
:+1:

PR #684, opened 2015-01-19
Print "verified" tag when updating Kendo UI

teobugslayer:
Also, pretty-print all the information in a nicely drawn table.

implements http://teampulse.telerik.com/view#item/283189


rosen-vladimirov:
:+1:

PR #685, opened 2015-01-19
Pretty-print iTunes connect application info for appstore list command

teobugslayer:
implements http://teampulse.telerik.com/view#item/280886


rosen-vladimirov:
:+1:

PR #687, opened 2015-01-19
Fix usage of shorthand options

rosen-vladimirov:
Fix usage of shorthands, so when you specify -t for example to really use specified template. Update common lib where shorthands for help command had been added.

http://teampulse.telerik.com/view#item/284030


teobugslayer:
:+1:

PR #688, opened 2015-01-19
Update to TypeScript 1.4

teobugslayer:
See https://github.com/telerik/mobile-cli-lib/pull/141


rosen-vladimirov:
:+1:  but after green build (we can stop other builds)

PR #693, opened 2015-01-20
Fix exception in getCommandName

rosen-vladimirov:
Remove ")" from the end of some unit tests names. Update common lib where the following change is applied:
Add toString() call before calling toLowerCase() to make sure the value is a string. This prevents exception when passed value is a number.

Fixes http://teampulse.telerik.com/view#item/284099


teobugslayer:
:+1:

PR #695, opened 2015-01-20
Turn on downloading of debugging tools & simulator from the CDN.

teobugslayer:
Download times went from 03:26 min to 00:17 min.


rosen-vladimirov:
:+1:

PR #700, opened 2015-01-21
Init command to extract .*.abproject and .abignore

rosen-vladimirov:
Add initializeProjectFromExistingFiles method that creates .abproject files and extracts .debug.abproject, .release.abproject and .abignore files from Blank template. If any of the files already exists, it will not be overwritten.

http://teampulse.telerik.com/view#item/284263

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -198,8 +198,7 @@ export class Project implements Project.IProject {
 		return this.createFromTemplate(projectName, projectDir);
 	}
 
-
-	public createProjectFileFromExistingProject(framework: string): IFuture<void> {
+	public initializeProjectFromExistingFiles(framework: string): IFuture<void> {
 		return ((): void => {
 			var projectDir = this.getNewProjectDir();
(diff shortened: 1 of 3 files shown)


lib/project.ts
```diff
@@ -212,7 +211,14 @@ export class Project implements Project.IProject {
 			}
 
 			this.frameworkProject = this.$frameworkProjectResolver.resolve(framework);
+			this.createProjectFileFromExistingProject(projectDir).wait();
+			var blankTemplateFile = this.frameworkProject.getTemplateFilename("Blank");
+			this.$fs.unzipSpecificFiles(path.join(this.$templatesService.projectTemplatesDir, blankTemplateFile), projectDir, [".*.abproject", ".abignore"], { overwriteExisitingFiles: false }).wait();
```
Fatme: Shouldn't we extract configuration specific files too?

Fatme:
:+1:

PR #704, opened 2015-01-21
Add support for nativescript 0.5.0

rosen-vladimirov:
Update range of supported nativescript versions to use 0.5.0 and add verification value in our unit tests to be 0.5.0.


Fatme:
:+1:

PR #705, opened 2015-01-22
Fix plugin find command

Mitko-Kerezov:
The root of this problem resides in
plugman's search function line 29 in ~/lib/services/cordova-plugins.ts
It doesn't really return IPlugin so a new interface is required

Related to http://teampulse.telerik.com/view#item/284370

### DIFF
--- a/lib/commands/plugin/find-plugins.ts
+++ b/lib/commands/plugin/find-plugins.ts
@@ -6,7 +6,8 @@ import util = require("util");
 import os = require("os");
 
 export class FindPluginsCommand implements ICommand {
-	constructor(private $cordovaPluginsService: service.CordovaPluginsService) {
+	constructor(private $cordovaPluginsService: service.CordovaPluginsService,
+				private $logger: ILogger) {
 	}
 
 	public allowedParameters: ICommandParameter[] = [];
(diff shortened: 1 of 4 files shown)


lib/commands/plugin/find-plugins.ts
```diff
@@ -24,17 +24,17 @@ export class FindPluginsCommand implements ICommand {
 		}).future<void>()();
 	}
 
-	private printPlugins(plugins: IPlugin[]) {
+	private printPlugins(plugins: IShortInformationPlugin[]) {
 		_.each(plugins, (plugin) => {
 			var pluginDescription = this.composePluginDescription(plugin);
 			console.log(pluginDescription);
```
rosen-vladimirov: I know this is not part of your changes, but could you replace this console.log with $logger.info. Also we may consider beautifying the output by using Table as @teobugslayer improved output of appstore list command. But for the moment I think using $logger.info will be enough.

lib/services/cordova-plugins.ts
```diff
@@ -19,13 +19,13 @@ export class CordovaPluginsService implements ICordovaPluginsService {
 		this.$project.ensureCordovaProject();
 	}
 
-	public getPlugins(keywords: string[]): IPlugin[] {
+	public getPlugins(keywords: string[]): IShortInformationPlugin[] {
```
rosen-vladimirov: Have you tested plugins fetch command after this change? I think it is using getPlugins method as well, so this will impact it as well.

Mitko-Kerezov: Yep - tested out plugins fetch command as well - works like a charm

rosen-vladimirov:
:+1:  I'm fine with the change. The only concern I have is about ```IShortInformationPlugin``` name, which could be replaced with IBasicInformationPlugin for example. But as I couldn't find more suitable name, you have my approval.

Mitko-Kerezov:
Yeah - I had trouble with the naming. 
@rosen-vladimirov I like your name better so I'm changing it

Fatme:
My suggestion is to use something like `IPlugmanPluginInformation`

PR #706, opened 2015-01-22
Improve error when unable to pass simulator params

rosen-vladimirov:
Add correct set of --plugins when starting simulator, in cases when CorePlugins are not defined.

Fixes http://teampulse.telerik.com/view#item/284384


ligaz:
:+1:

PR #708, opened 2015-01-23
Trigger cordova migration when $appbuilder prop set FrameworkVersion is executed

Fatme:


rosen-vladimirov:
:+1:  after red build with 1-2 failing tests :D

PR #712, opened 2015-01-26
Switch to external 7zip binary to uncompress files.

teobugslayer:
Switch to external 7zip binary to uncompress files. This supports both zip and lzma compression methods.

See http://teampulse.telerik.com/view#item/284635


rosen-vladimirov:
:+1:

PR #714, opened 2015-01-26
Fixes the exception that is thrown when initializing project in empty directory

Fatme:
http://teampulse.telerik.com/view#item/284263


rosen-vladimirov:
:+1:

PR #715, opened 2015-01-26
Add --validValue option to prop print command

rosen-vladimirov:
Add validValue to options. When it is specified, all valid values of the requested property will be printed. Improve prop print command to work in the following way:
I. inside project dir:
     1. $ appbuilder prop print -> prints all properties of current project and their current values
     2. $ appbuilder prop print --validValue -> prints all properties valid for the current project and all valid values for each of them.
     3. $ appbuilder prop print <PropName> -> prints the value of the property for the current project
     4. $ appbuilder prop print <PropName> --validValue -> prints all valid values for the selected property, applicable to the current project type.

II. Outside of project dir:
     1. $ appbuilder prop print -> leads to error, that no project is found
     2. $ appbuilder prop print --validValue -> prints all properties for all projects and all valid values for each of them.
     3. $ appbuilder prop print <PropName> -> leads to error, that no project is found
     4. $ appbuilder prop print <PropName> --validValue -> prints all valid values for the selected property

Remove dynamic call from help.txt in order to make the output more readable. Remove calls to ensureProject when working with prop print command.

http://teampulse.telerik.com/view#item/284655

### DIFF
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -195,6 +195,8 @@ interface IProjectPropertiesService {
 	completeProjectProperties(properties: any, frameworkProject: Project.IFrameworkProject): boolean;
 	updateProjectProperty(projectData: any, mode: string, property: string, newValue: any): IFuture<void>;
 	normalizePropertyName(property: string, projectData: IProjectData): string;
+	getValidValuesForProperty(propData: any): IFuture<string[]>;
+	getPropertiesForAllSupportedProjects(): IFuture<string>;
 }
 
 interface IServerConfigurationData {
--- a/lib/json-schema/json-schema-validator.ts
+++ b/lib/json-schema/json-schema-validator.ts
@@ -66,16 +66,17 @@ export class JsonSchemaValidator implements IJsonSchemaValidator {
 					result[key] = schema.extends.properties[key];
 				}
 			});
-			var projectPropertiesFilePath = this.$resources.resolvePath(util.format("project-properties-%s.json",framework.toLowerCase()));
-			if(this.$fs.exists(projectPropertiesFilePath).wait()) {
-				var fileContent = this.$fs.readJson(projectPropertiesFilePath).wait();
-				var additionalProperties = _.keys(fileContent);
-				_.each(additionalProperties, (propertyName: string) => {
-					_.each(_.keys(fileContent[propertyName]), (value: string) => {
-						result[propertyName][value] = fileContent[propertyName][value];
-					});
+		}
+
+		var projectPropertiesFilePath = this.$resources.resolvePath(util.format("project-properties-%s.json",framework.toLowerCase()));
+		if(this.$fs.exists(projectPropertiesFilePath).wait()) {
+			var fileContent = this.$fs.readJson(projectPropertiesFilePath).wait();
+			var additionalProperties = _.keys(fileContent);
+			_.each(additionalProperties, (propertyName: string) => {
+				_.each(_.keys(fileContent[propertyName]), (key: string) => {
+					result[propertyName][key] = fileContent[propertyName][key];
 				});
-			}
+			});
 		}
 
 		return result;
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -22,7 +22,9 @@ var knownOpts: any = {
 		"client": String,
 		"available": Boolean,
 		"release": Boolean,
-		"debug": Boolean
+		"debug": Boolean,
+		"valid-value": Boolean,
+		"validValue": Boolean
 	},
 	shorthands: IStringDictionary = {
 		"t": "template",
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -16,6 +16,7 @@ export class Project implements Project.IProject {
 	private static CONFIGURATION_FILE_SEARCH_PATTERN: RegExp = new RegExp(".*.abproject$", "i");
 	private static VALID_CONFIGURATION_CHARACTERS_REGEX = "[-_A-Za-z0-9]";
 	private static CONFIGURATION_FROM_FILE_NAME_REGEX = new RegExp("^[.](" + Project.VALID_CONFIGURATION_CHARACTERS_REGEX + "+?)" + Project.JSON_PROJECT_FILE_NAME_REGEX + "$", "i");
+	private static INDENTATION = "     ";
 
 	private _hasBuildConfigurations: boolean = false;
 	private _projectSchema: any;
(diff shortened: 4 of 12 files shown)


lib/project/project.d.ts
```diff
@@ -44,8 +44,6 @@ declare module Project {
 		requiredAndroidApiLevel: number;
 		configFiles: IConfigurationFile[];
 		getTemplateFilename(name: string): string;
-		projectTemplatesString(): IFuture<string>;
-		alterPropertiesForNewProject(properties: any, projectName: string): void;
```
rosen-vladimirov: Removed as they are duplicated :)

rosen-vladimirov:
ping @ikoevska  for the help.txt - we have to improve the information about prop print command. I've written just the base idea, but maybe we should include more information how the command works in different situatuions (check PR's description or teampulse item for more info).

lib/json-schema/json-schema-validator.ts
```diff
@@ -63,18 +63,19 @@ export class JsonSchemaValidator implements IJsonSchemaValidator {
...
-				});
-			}
 		}
 
+        var projectPropertiesFilePath = this.$resources.resolvePath(util.format("project-properties-%s.json",framework.toLowerCase()));
+        if(this.$fs.exists(projectPropertiesFilePath).wait()) {
+            var fileContent = this.$fs.readJson(projectPropertiesFilePath).wait();
+            var additionalProperties = _.keys(fileContent);
+            _.each(additionalProperties, (propertyName: string) => {
+                _.each(_.keys(fileContent[propertyName]), (value: string) => {
```
teobugslayer: We read the keys, but the callback receives values? Which is which?

lib/declarations.d.ts
```diff
@@ -199,6 +199,14 @@ interface IProjectPropertiesService {
 	completeProjectProperties(properties: any, frameworkProject: Project.IFrameworkProject): boolean;
 	updateProjectProperty(projectData: any, mode: string, property: string, newValue: any): IFuture<void>;
 	normalizePropertyName(property: string, projectData: IProjectData): string;
+	getValidValuesForProperty(propData: any): IFuture<any[]>
+	getPropertiesForAllSupportedProjects(): IFuture<string>;
+}
+
+interface IPropertyValue {
+	id: string;
```
teobugslayer: Is this member really needed?

lib/options.ts
```diff
@@ -22,7 +22,8 @@ var knownOpts: any = {
 		"client": String,
 		"available": Boolean,
 		"release": Boolean,
-		"debug": Boolean
+		"debug": Boolean,
+		"validValue": Boolean
```
teobugslayer: will $ appbuilder blah --validvalue (note the lower case) work on Windows?

lib/project.ts
```diff
@@ -358,23 +359,77 @@ export class Project implements Project.IProject {
...
+						_.each(_.keys(schema), propKey => {
+							var prop = schema[propKey];
+							this.$logger.info("  " + propKey);
+							this.printValidValuesOfProperty(prop).wait();
+						});
+					} else {
+						// 'appbuilder prop print' called inside project dir
+						var propKeys: any = _.keys(this.projectData);
+						var sortedProperties = _.sortBy(propKeys, (propertyName: string) => {
+							return propertyName.toLowerCase();
```
teobugslayer: Maybe we can move toLowerCase outside sorting? for increased efficiency ;)

lib/project.ts
```diff
@@ -358,23 +359,77 @@ export class Project implements Project.IProject {
...
+						if(_.has(this.projectData, normalizedPropertyName)) {
+							this.$logger.out(this.projectData[normalizedPropertyName]);
+						} else {
+							this.$errors.fail("Unrecognized project property '%s'", property);
+						}
+					}
+				} else {
+					if(options.validValue) {
+						// 'appbuilder prop print --validValue' called inside project dir
+						_.each(_.keys(schema), propKey => {
```
rosen-vladimirov: sort the keys

lib/options.ts
rosen-vladimirov: no, we have two options here: add valid-value as known option or change validValue to be only valid.
Ping @Fatme @ikoevska  - which one do you think is better?

lib/project.ts
```diff
@@ -358,23 +359,77 @@ export class Project implements Project.IProject {
...
+			if(property.pattern) {
+				this.$logger.trace("%sDesired pattern is: %s", Project.INDENTATION, property.pattern);
 			}
 
+			var validValues: IPropertyValue[] = this.$projectPropertiesService.getValidValuesForProperty(property).wait();
+			if(validValues && validValues.length > 0) {
+				this.$logger.out("%sValid values:", Project.INDENTATION);
+			}
+			_.forEach(validValues, (val: any) => {
+				this.$logger.out("%s  %s    %s", Project.INDENTATION, val.id, val.description);
```
teobugslayer: We can require("table") here and pretty-print it.

lib/services/project-properties-service.ts
```diff
@@ -162,6 +168,21 @@ export class ProjectPropertiesService implements IProjectPropertiesService {
 		}).future<string[]>()();
 	}
 
+	public getValidValuesForProperty(propData: any): IFuture<IPropertyValue[]> {
+		return ((): IPropertyValue[]=> {
+			var validValues: PropertyValue[] = [];
+			var range = this.getPropRange(propData).wait();
+			if(range) {
+				_.each(range, (rangeDesc: any, rangeKey: any) => {
```
teobugslayer: Why not _.map range -> validValues?

lib/project.ts
Fatme: It seems that somethimes `val.description` is `undefined` and we print `undefined` on the console. I got the following output 
`Framework
     Valid values:
       Cordova    undefined` when I execute `appbuilder prop print --validValue`. I suggest to print the `description`  only when we have such.

lib/project.ts
rosen-vladimirov: Well, the idea is to persist the prioperties as they are, but to sort them based on their toLowerCase values. That's why toLowerCase is inside the sortBy

lib/project.ts
```diff
@@ -363,23 +364,82 @@ export class Project implements Project.IProject {
...
-				this.$errors.fail("Unrecognized project property '%s'", property);
+			if(property.pattern) {
+				this.$logger.trace("%sDesired pattern is: %s", Project.INDENTATION, property.pattern);
 			}
 
+			var validValues: string[] = this.$projectPropertiesService.getValidValuesForProperty(property).wait();
+			if(validValues) {
+				this.$logger.out("%sValid values:", Project.INDENTATION);
+			}
+			_.forEach(validValues, value => {
```
rosen-vladimirov: maybe move the _.forEach inside the above if

Fatme:
:+1:

teobugslayer:
After discussing in person, :+1:

PR #717, opened 2015-01-26
Don't trigger json schema validation for prop commands

Fatme:


teobugslayer:
:+1:  after green build

lib/declarations.d.ts
```diff
@@ -230,6 +230,7 @@ interface IStaticConfig extends Config.IStaticConfig {
 	SOLUTION_SPACE_NAME: string;
 	DEBUG_PROJECT_FILE_NAME: string;
 	RELEASE_PROJECT_FILE_NAME: string;
+	triggerJsonSchemaValidation: boolean;
```
teobugslayer: Can we shorten the name a bit? I would suggest something similar to "validateProject" or "validateProjectSchema"

rosen-vladimirov:
:+1:

PR #721, opened 2015-01-26
Fix build after moving opener from AB CLI into the common lib

teobugslayer:


rosen-vladimirov:
:+1:  after green build

lib/declarations.d.ts
```diff
@@ -162,7 +162,7 @@ interface IProjectCapabilities {
 }
 
 interface IOpener {
-	open(filename: string): void;
+	open(filename: string, appname?: string): void;
```
Fatme: Shouldn't this interface be in common lib?

Fatme:
You should also remove opener from bootstrap.

lib/declarations.d.ts
teobugslayer: Done! Nice catch!

PR #722, opened 2015-01-27
Fix plugin fetch command

Mitko-Kerezov:
Fixes http://teampulse.telerik.com/view#item/284717

### DIFF
--- a/lib/commands/plugin/fetch-plugin.ts
+++ b/lib/commands/plugin/fetch-plugin.ts
@@ -22,13 +22,14 @@ export class FetchPluginCommand implements ICommand {
 				this.$logger.out(result);
 			} else {
 				var plugins = this.$cordovaPluginsService.getPlugins(args);
-				var pluginsCount = Object.keys(plugins).length;
+				var pluginKeys = Object.keys(plugins);
+				var pluginsCount = pluginKeys.length;
 				if (pluginsCount === 0) {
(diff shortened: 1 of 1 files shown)


lib/commands/plugin/fetch-plugin.ts
```diff
@@ -22,13 +22,14 @@ export class FetchPluginCommand implements ICommand {
 				this.$logger.out(result);
 			} else {
 				var plugins = this.$cordovaPluginsService.getPlugins(args);
-				var pluginsCount = Object.keys(plugins).length;
+				var pluginKeys = Object.keys(plugins);
+				var pluginsCount = pluginKeys.length;
 				if (pluginsCount === 0) {
 					this.$logger.out("There are 0 matching plugins.");
-				} else if (pluginsCount > 1) {
+				} else if (pluginsCount > 1 && pluginKeys[0] != args[0]) {
```
rosen-vladimirov: !==
also are you sure that the plugins are sorted, for example if you have:
plugin, plugin1, plugin2
Are you sure they'll always be in this order? In case they are plugin1, plugin, plugin2 and you search for plugin, this check will fail.

Mitko-Kerezov:
@rosen-vladimirov Boolean operator fixed. As for the ordering - plugman always returns the plugins alphabetically sorted so plugin will always be before plugin1

rosen-vladimirov:
great
:+1:  after green build

PR #725, opened 2015-01-27
Allow only Cordova projects to be published in the appstore

teobugslayer:
Improve the error message when non-Corodova project is used.

Fixes http://teampulse.telerik.com/view#item/283182


rosen-vladimirov:
:+1:

PR #727, opened 2015-01-27
Remove Friends template test

rosen-vladimirov:
As Friends template is no longer downloaded, we must remove the test that's checking its files in order to fix the build.

http://teampulse.telerik.com/view#item/284783


teobugslayer:
:+1:

PR #728, opened 2015-01-27
Fix checking for Android SDK

teobugslayer:
Fix error handling on process spawn

Fixes http://teampulse.telerik.com/view#item/279833


rosen-vladimirov:
:+1:

PR #731, opened 2015-01-27
Use new api for chunk upload

rosen-vladimirov:
Use new api for chunk upload, where duplicate name (importProject) has been removed, so new one is used - importLocalProject. In order to do this I have:
 - fixed swagger generation when api has slash
 - regenerated server api
 - replace call to importProject1 with call to importLocalProject in MultipartUploadService

http://teampulse.telerik.com/view#item/284838


teobugslayer:
:+1:

PR #732, opened 2015-01-27
Fix error when trying to show the help for the create website command

teobugslayer:
The bug is the wrong case of a single symbol, which prevents dynamic calls from working


rosen-vladimirov:
:+1:

PR #734, opened 2015-01-28
Update TAM Publishing

rosen-vladimirov:
Uploading application in AppManager now uses the following rules for certificates/provisions:
 - android upload should work no matter of the certificate. If user had not specified certificate, we'll print warning and we'll use server's debug certificate
 - ios upload cannot use AppStore provisioning (Development, AdHoc and Enterprise are allowed)

http://teampulse.telerik.com/view#item/284661


teobugslayer:
:+1:

PR #735, opened 2015-01-28
Fix init command

Fatme:
When initialising new project we set FrameworkVersion to 3.2.0 for cordova projects and to 1.0.0 for NativeScript projects because we read them from default-project files. I removed these files. We extract Blank template to add .abignore and configuration specific files. I added .abproject file from Blank template and use its project properties.


lib/project.ts
```diff
@@ -223,24 +223,23 @@ export class Project implements Project.IProject {
 			}
 
 			this.frameworkProject = this.$frameworkProjectResolver.resolve(framework);
-			this.createProjectFileFromExistingProject(projectDir, framework).wait();
 			var blankTemplateFile = this.frameworkProject.getTemplateFilename("Blank");
-			this.$fs.unzip(path.join(this.$templatesService.projectTemplatesDir, blankTemplateFile), projectDir, { overwriteExisitingFiles: false }, [".*.abproject", ".abignore"]).wait();
+			this.$fs.unzip(path.join(this.$templatesService.projectTemplatesDir, blankTemplateFile), projectDir, { overwriteExisitingFiles: false }, [".*.abproject", ".abignore", ".abproject"]).wait();
```
rosen-vladimirov: just *.abproject should be enough instead of .*.abproject and .abproject

lib/services/project-properties-service.ts
```diff
@@ -40,16 +40,6 @@ export class ProjectPropertiesService implements IProjectPropertiesService {
 
 		updated = frameworkProject.completeProjectProperties(properties);
```
rosen-vladimirov: not part of the current changes, but I've noticed that before this line you have one if, in which you are setting updated = true; In case completeProjectProperties returns false, you'll overwrite this updated value

lib/project.ts
rosen-vladimirov: I've tested it, you can use 
```Typescript
this.$fs.unzip(path.join(this.$templatesService.projectTemplatesDir, blankTemplateFile), projectDir, { overwriteExisitingFiles: false }, ["*.abproject", ".abignore"]).wait();
```

rosen-vladimirov:
:+1: one minor comment and one suggestion, not relevant to this PR

PR #737, opened 2015-01-28
Update minimatch to the latest version

teobugslayer:
See http://teampulse.telerik.com/view#item/284931


rosen-vladimirov:
:+1:  after green build

PR #740, opened 2015-01-29
Fix using of properties from configuration specific .abproject

rosen-vladimirov:
In case you have CorePlugins defined in .abproject only (not in .debug.abproject and .release.abproject) you cannot build your project. Add check if the property is defined in the configuration specific files and if not - check for it in .abproject.

Fixes http://teampulse.telerik.com/view#item/284955


Fatme:
:+1:

PR #745, opened 2015-01-29
Remove password prompt when uploading .cer certificate

Mitko-Kerezov:
Those come from Apple and don't have a password
Fixes http://teampulse.telerik.com/view#item/273404

### DIFF
--- a/lib/commands/cryptographic-identities.ts
+++ b/lib/commands/cryptographic-identities.ts
@@ -15,6 +15,8 @@ import commandParams = require("../common/command-params");
 class CryptographicIdentityConstants {
 	public static PKCS12_EXTENSION = "p12";
 	public static X509_EXTENSION = "cer";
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
:+1:  after green build

lib/commands/cryptographic-identities.ts
```diff
@@ -590,7 +590,7 @@ export class ImportCryptographicIdentity implements ICommand {
 				this.$errors.fail("The file '%s' does not exist.", certificateFile);
 			}
 
-			if(!password) {
+			if(!password && importType === "Pkcs12") {
```
Fatme: I suggest you to extract `Pkcs12` as a constant. We use it more than one time.

Mitko-Kerezov:
Addressed comment by @Fatme

Fatme:
:+1:

PR #746, opened 2015-01-29
Batch LiveSync over cable

teobugslayer:
Keep all changes and hit the device every 500 ms

See http://teampulse.telerik.com/view#item/285009


rosen-vladimirov:
:+1:

Fatme:
:+1:

PR #748, opened 2015-01-29
Add check if 7za can be executed on Linux

rosen-vladimirov:
If user is on x64 Linux and 32 bit libraries, describe in our documentation, are not installed, user will not be able to use 7zip. This means - unable to create projects, unable to build etc. So on post-install we check if we are able to execute the 7za file and if ENOENT is return we show the user where he can find system requirements.

Fix host-info paths in some files and remove host-info reference from some other files.

Remove duplicated methods from host-info files (they exist on two places - in common lib and here).

http://teampulse.telerik.com/view#item/285027

### DIFF
--- a/lib/host-info.ts
+++ b/lib/host-info.ts
@@ -2,26 +2,6 @@
 "use strict";
 import Future = require("fibers/future");
 
-export function isWindows() {
-	return process.platform === "win32";
-}
-
-export function isWindows64() {
-	return isWindows() && (process.arch === "x64" || process.env.hasOwnProperty("PROCESSOR_ARCHITEW6432"));
-}
-
-export function isWindows32() {
-	return isWindows() && !isWindows64();
-}
-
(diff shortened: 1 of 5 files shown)


teobugslayer:
:+1: after addressing the issues in https://github.com/telerik/mobile-cli-lib/pull/162/files

lib/host-info.ts
```diff
@@ -38,8 +18,8 @@ export function isDotNet40Installed(message: string) : IFuture<boolean> {
 	return result;
 }
 
-export var hostCapabilities: { [key:string]: IHostCapabilities } = {
-	"win32": { 
+export var hostCapabilities: { [key: string]: IHostCapabilities } = {
```
teobugslayer: Maybe IDictionary\<IHostCapabilities\>?

rosen-vladimirov:
closing and will open against release branch

PR #751, opened 2015-01-30
Fix error when uploading large projects

rosen-vladimirov:
We start all chunk uploads simultaneously as fibers are started on process.nextTick. Limit them to three simultaneously by waiting the batch of three before creating the next three chunks. Add message to indicate uploading project to server.

It is part of http://teampulse.telerik.com/view#item/283781

### DIFF
--- a/lib/services/build.ts
+++ b/lib/services/build.ts
@@ -71,9 +71,7 @@ export class BuildService implements Project.IBuildService {
 			buildProperties.LiveSyncToken = liveSyncToken;
 
 			var buildProjectFuture = this.$server.build.buildProject(solutionName, projectName, { Properties: buildProperties, Targets: [] });
-			while(!buildProjectFuture.isResolved()) {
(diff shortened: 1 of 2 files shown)


lib/services/build.ts
```diff
@@ -496,17 +494,30 @@ export class BuildService implements Project.IBuildService {
...
+				this.printDotsWhileWaiting(this.$server.projects.importProject(projectName, projectName,
+					this.$fs.createReadStream(projectZipFile)), 2000).wait();
 			}
+
+			commonHelpers.printInfoMessageOnSameLine(os.EOL);
 			this.$logger.trace("Project imported");
 		}).future<void>()();
 	}
+
+	private printDotsWhileWaiting(future: IFuture<any>, timeout: number): IFuture<void> {
```
teobugslayer: I would change the name to something like "printProgressIndicator" or smthg in the same vein.

teobugslayer:
A minor recommendation, otherwise :+1:

PR #760, opened 2015-02-02
Fix appmanager upload ios with specified provision

rosen-vladimirov:
In case you specify --provision, an error is thrown. Replace incorrect Server.ProvisionType.AppStore with constants.ProvisionType.AppStore

http://teampulse.telerik.com/view#item/285210


teobugslayer:
:+1:

PR #761, opened 2015-02-02
Respect options.release from emulate command

Fatme:
Fixes http://teampulse.telerik.com/view#item/285220


rosen-vladimirov:
:+1:

PR #762, opened 2015-02-02
Improve .abproject prop completion

rosen-vladimirov:
Move completion of properties before validation of the project in order to have correct WP8 props.

If .abproject is missing projectVersion we print warning and add it.

In old project instead of Framework, we had projectType - in case we find projectType in .abproject, replace it with Framework.

http://teampulse.telerik.com/view#item/285225

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -578,18 +578,21 @@ export class Project implements Project.IProject {
 	private readProjectData(): IFuture<void> {
 		return (() => {
 			var projectDir = this.getProjectDir().wait();
-
+			var shouldSaveProject = false;
 			if(projectDir) {
 				var projectFilePath = path.join(projectDir, this.$staticConfig.PROJECT_FILE_NAME);
 				try {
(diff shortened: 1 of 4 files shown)


lib/project.ts
```diff
@@ -578,18 +578,21 @@ export class Project implements Project.IProject {
...
 		return (() => {
 			var projectDir = this.getProjectDir().wait();
-
+			var shouldSaveProject = false;
 			if(projectDir) {
 				var projectFilePath = path.join(projectDir, this.$staticConfig.PROJECT_FILE_NAME);
 				try {
 					var data = this.$fs.readJson(projectFilePath).wait();
-
+
```
teobugslayer: Can we remove this empty space, please?

teobugslayer:
:+1:

rosen-vladimirov:
This PR should be reworked, currently unit tests are failing, I'm working on it, so don't look at it for the moment

teobugslayer:
:+1:

PR #769, opened 2015-02-03
Fix genymotion check

rosen-vladimirov:
Update common lib where the following change is applied:
When you specify --geny option of $ appbuilder emulate android command, we check for genyshell in the PATH environment variable. As on MacOS genyshell and the real program that we need for genymotion (player) are in different directories, we fail with incorrect message.

Fixes http://teampulse.telerik.com/view#item/285302


nmladenov:
:+1:

Fatme:
:+1:

PR #771, opened 2015-02-03
Print plugin variables value's when executing appbuilder plugin

Fatme:
Fixes http://teampulse.telerik.com/view#item/281723

### DIFF
--- a/lib/plugins-data.ts
+++ b/lib/plugins-data.ts
@@ -1,6 +1,7 @@
 ///<reference path=".d.ts"/>
 "use strict";
 
+import os = require("os");
 import util = require("util");
 
 export enum PluginType {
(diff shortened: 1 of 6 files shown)


lib/plugins-data.ts
```diff
@@ -13,56 +14,88 @@ export class CordovaPluginData implements IPlugin {
...
+		return result;
+	}
+
+	private getBasicPluginInformation(): string[] {
+		var nameRow = this.buildRow("Plugin", this.data.Name);
+		var identifierRow = this.buildRow("Identifier", this.data.Identifier);
+		var versionRow = this.buildRow("Version", this.data.Version);
+		var urlRow = this.buildRow("Url", this.data.Url);
+
+		 var result = [nameRow, identifierRow, versionRow, urlRow];
```
rosen-vladimirov: space befor var (don't hate me too much)

lib/plugins-data.ts
```diff
@@ -13,56 +14,88 @@ export class CordovaPluginData implements IPlugin {
...
+		return util.format("    %s: %s", key, value);
+	}
+
+	protected composePluginInformation(additionalPluginData: string[]) 	{
+		var result = this.getBasicPluginInformation();
+
+		result = result.concat(additionalPluginData);
+		result = result.concat(this.getPluginVariablesInfo());
+
+		return result;
```
teobugslayer: maybe ``` return _.flatten(this.getBasicPluginInformation(), additionalPluginData, this.getPluginVariablesInfo())```?

lib/plugins-data.ts
```diff
@@ -13,56 +14,88 @@ export class CordovaPluginData implements IPlugin {
...
+				var variables = pluginVariablesData[this.data.Identifier];
+				var variableNames = _.keys(variables);
+				if(variableNames.length > 0) {
+					var output:string[] = [];
+					output.push(util.format("    Variables for %s configuration: %s", configuration, os.EOL));
+					_.each(variableNames, (variableName:string) => {
+						output.push(util.format("        %s: %s %s", variableName, variables[variableName], os.EOL));
+					});
+
+					result.push(output.join(""));
```
teobugslayer: Why not ```output.join(os.EOL)``` and remove it from the other places?

teobugslayer:
:+1:

PR #774, opened 2015-02-03
Fix run command

Fatme:
Fixes http://teampulse.telerik.com/view#item/283764
Should be merged after this PR https://github.com/telerik/mobile-cli-lib/pull/168


rosen-vladimirov:
:+1:  after rebasing

PR #775, opened 2015-02-03
Show plugin url for marketplace plugins

Fatme:
Fixes http://teampulse.telerik.com/view#item/281518
Sometimes server returns empty url address but we actually have the url from the response from `telerik.plugins.com`. This is only temporaly fix. After integrating the feature for marketplace plugin versioning we should delete this code.


rosen-vladimirov:
:+1:

PR #776, opened 2015-02-04
Add verified publisher for marketplace plugins

rosen-vladimirov:
If plugin is verified its publisher is Telerik or Telerik Partner. When showing info for plugins, if it is verified we show Telerik, or Telerik Partner.  Add IPublisher interface and modify MarketplacePluginData to handle verified plugins.

Fixes http://teampulse.telerik.com/view#item/280939


teobugslayer:
:+1: (as verified by grepping on Mac)

PR #780, opened 2015-02-04
Limit node version to be below 0.11.00

rosen-vladimirov:
Limit node version to be below 0.11.00. We need this as iojs is recognized as node 1.1.0 (or above) and we cannot work with it. The limit will cause only warning when installing appbuilder.

http://teampulse.telerik.com/view#item/285415


teobugslayer:
:+1:

PR #782, opened 2015-02-05
Fix emulate of NativeScript projects

rosen-vladimirov:
Fix getProjectTargets of nativescript-project to be "android" and "ios" (remove code which was checking for bootstrap.android.js and bootstrap.ios.js). Fix emulate command to use appIdentifier directly (without using AppIdentifier class and passing incorrect options.companion option to it). Remove projectTargets property of project, as there is getProjectTargets method, which is used almost everywhere when we need project targets (fixed the only place where the projectTargets property was used).  Add emulate option to project capabilities in order to forbid emulate of mobile-website projects. EmulatorSettingsService checks for emulate capability and fails in case the project cannot be emulated.

NOTE: This PR allows to deploy nativescript projects to native emulator. For Android, the application is installed, but it is not started. It has to be started manually. Graphite is doing the same, I'll research the issue and if I find a fix, I'll include it in another PR.

Fixes http://teampulse.telerik.com/view#item/285409

### DIFF
--- a/lib/project/nativescript-project.ts
+++ b/lib/project/nativescript-project.ts
@@ -3,6 +3,7 @@
 
 import path = require("path");
 import util = require("util");
+import Future = require("fibers/future");
 
 import frameworkProjectBaseLib = require("./framework-project-base");
 import helpers = require("./../common/helpers");
@@ -34,7 +35,8 @@ export class NativeScriptProject extends frameworkProjectBaseLib.FrameworkProjec
 			simulate: false,
 			livesync: false,
 			livesyncCompanion: true,
-			updateKendo: false
+			updateKendo: false,
+			emulate: true
 		};
 	}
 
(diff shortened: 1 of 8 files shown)


lib/project/nativescript-project.ts
```diff
@@ -79,10 +80,9 @@ export class NativeScriptProject extends frameworkProjectBaseLib.FrameworkProjec
 	}
 
 	public getProjectTargets(projectDir: string): IFuture<string[]> {
-		var dir = path.join(projectDir, "app");
-		var fileMask = /^bootstrap\.(\w*)\.js$/i;
-
-		return this.getProjectTargetsBase(dir, fileMask);
+		return (() => {
```
teobugslayer: You can use Future.fromResult to shorten the code.

lib/services/emulator-settings-service.ts
```diff
@@ -1,13 +1,22 @@
...
-	constructor(private $project: Project.IProject) { }
+	constructor(private $project: Project.IProject,
+		private $errors: IErrors) { }
 
 	public canStart(platform: string): IFuture<boolean> {
 		return (() => {
 			this.$project.ensureProject();
-			return _.contains(this.$project.projectTargets.wait(), platform.toLowerCase());
+
+			this.$project.capabilities.emulate
```
teobugslayer: Probably a debug code?

lib/services/emulator-settings-service.ts
```diff
@@ -1,13 +1,22 @@
 ///<reference path="../.d.ts"/>
 "use strict";
 
+var util = require("util");
```
rosen-vladimirov: remove it as it is not used anywhere

teobugslayer:
:+1:

tytodorov:
:+1:

PR #784, opened 2015-02-05
Fix projectVersion completion

rosen-vladimirov:
If projectVersion is missing from .abproject, we should set it to 1. Fix the code to set it to number 1, not string 1.

http://teampulse.telerik.com/view#item/285484


teobugslayer:
:+1:

tytodorov:
:+1:

PR #787, opened 2015-02-06
Print configuration specific properties

rosen-vladimirov:
"$ appbuilder prop print" and "$ appbuilder prop print CorePlugins" do not print the values of the configuration specific properties. Use cli-table to print the values of all configuration specific properties for all configurations (it will work no matter how much configurations we add in the future).

Fixes http://teampulse.telerik.com/view#item/285533

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -428,6 +428,8 @@ export class Project implements Project.IProject {
 						// '$ appbuilder prop print <PropName>' called inside project dir
 						if(_.has(this.projectData, normalizedPropertyName)) {
 							this.$logger.out(this.projectData[normalizedPropertyName]);
+						} else if(this.hasConfigurationSpecificDataForProperty(normalizedPropertyName)) {
+							this.printConfigurationSpecificDataForProperty(normalizedPropertyName);
 						} else {
 							this.$errors.fail("Unrecognized project property '%s'", property);
 						}
@@ -447,6 +449,7 @@ export class Project implements Project.IProject {
 						var propKeys: any = _.keys(this.projectData);
 						var sortedProperties = _.sortBy(propKeys, (propertyName: string) => propertyName.toUpperCase());
 						_.each(sortedProperties, (propertyName: string) => this.$logger.out(propertyName + ": " + this.projectData[propertyName]));
+						this.printConfigurationSpecificData();
 					}
 				}
 			} else {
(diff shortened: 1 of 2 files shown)


lib/project.ts
```diff
@@ -488,6 +492,78 @@ export class Project implements Project.IProject {
...
+			properties.push(_.keys(this.configurationSpecificData[config]));
+		});
+
+		properties = _.uniq(_.flatten(properties));
+
+		if(properties.length > 0) {
+			this.$logger.out("%sConfiguration specific properties: ", os.EOL);
+		}
+
+		_.forEach(properties, (property: string) => {
```
rosen-vladimirov: move this in the if above (again)

lib/project.ts
```diff
@@ -488,6 +492,78 @@ export class Project implements Project.IProject {
...
+
+		var sharedValues = _.intersection.apply(null, configsDataForProperty);
+		_.forEach(sharedValues, value => table.push([value, value]));
+
+		var configSpecificValues: any = [];
+		_.forEach(configsDataForProperty, (config: any) => {
+			configSpecificValues.push(_.difference(config, sharedValues));
+		});
+
+		var maxLength = _.max(_.values(configSpecificValues).map(value => {
```
teobugslayer: you can remove the { return }

lib/project.ts
```diff
@@ -488,6 +492,78 @@ export class Project implements Project.IProject {
...
+			return value.length;
+		}));
+
+		var numberOfConfigs = configSpecificValues.length;
+		for(var valueIndex = 0; valueIndex < maxLength; valueIndex++) {
+			var tempArray: any[] = [];
+			for(var configNum = 0; configNum < numberOfConfigs; configNum++) {
+				var value = configSpecificValues[configNum][valueIndex];
+				tempArray.push(value ? value : "");
```
teobugslayer: value || ""

lib/project.ts
```diff
@@ -488,6 +492,78 @@ export class Project implements Project.IProject {
...
+				result = true;
+				// break forEach
+				return false;
+			}
+		});
+
+		return result;
+	}
+
+	private printConfigurationSpecificDataForProperty(normalizedPropertyName: string): void {
```
teobugslayer: This function should either return the cli-table or even better, the array containing the output data. The caller function should be responsible for printing the data out.

lib/project.ts
```diff
@@ -488,6 +492,78 @@ export class Project implements Project.IProject {
...
+			configsDataForProperty.push(isArray ? data : [data]);
+		});
+
+		var table = new Table({
+			head: headers,
+			chars: { 'mid': '', 'left-mid': '', 'mid-mid': '', 'right-mid': '' }
+		});
+
+		var sharedValues = _.intersection.apply(null, configsDataForProperty);
+		_.forEach(sharedValues, value => table.push([value, value]));
```
rosen-vladimirov: this is not correct - what if you have only one configuration or more than 2?

lib/project.ts
```diff
@@ -488,6 +492,78 @@ export class Project implements Project.IProject {
...
+		}
+
+		this.$logger.out(table.toString());
+	}
+
+	private printConfigurationSpecificData(): void {
+		var properties: any[] = [];
+		_.forEach(_.keys(this.configurationSpecificData), config => {
+			properties.push(_.keys(this.configurationSpecificData[config]));
+		});
```
teobugslayer:
```
var properties: any[] = _(this.configurationSpecificData)
	.keys()
	.map(config => _.keys(this.configurationSpecificData[config]))
	.value();
```

This removes side effects and reads from left to right.

lib/project.ts
```diff
@@ -488,6 +492,78 @@ export class Project implements Project.IProject {
...
+
+		var maxLength = _.max(_.values(configSpecificValues).map(value => {
+			return value.length;
+		}));
+
+		var numberOfConfigs = configSpecificValues.length;
+		for(var valueIndex = 0; valueIndex < maxLength; valueIndex++) {
+			var tempArray: any[] = [];
+			for(var configNum = 0; configNum < numberOfConfigs; configNum++) {
```
teobugslayer: Is there a reason not to use _.each()?

lib/project.ts
```diff
@@ -488,6 +492,78 @@ export class Project implements Project.IProject {
 		}).future<void>()();
 	}
 
+	private hasConfigurationSpecificDataForProperty(normalizedPropertyName: string) {
+		var result = false;
+		_.forEach(_.keys(this.configurationSpecificData), (config: string) => {
```
teobugslayer: Can we use _.some() here?

lib/project.ts
```diff
@@ -488,6 +491,62 @@ export class Project implements Project.IProject {
...
+		return _.some(properties, prop => prop === normalizedPropertyName);
+	}
+
+	private getConfigurationSpecificDataForProperty(normalizedPropertyName: string): any[] {
+		var configsDataForProperty: any[] = [];
+		var numberOfConfigs = _.keys(this.configurationSpecificData).length;
+
+		_.forEach(_.keys(this.configurationSpecificData), config => {
+			var data = this.configurationSpecificData[config][normalizedPropertyName];
+			configsDataForProperty.push(_.flatten([data]));
```
teobugslayer: This is :cry:ing for a _.map

lib/project.ts
```diff
@@ -488,6 +491,62 @@ export class Project implements Project.IProject {
...
+
+		var sharedValues: string[] = _.intersection.apply(null, configsDataForProperty);
+		var valuesInAllConfigs = _.map(sharedValues, value => helpers.fill(value, numberOfConfigs));
+		var configSpecificValues = _.map(configsDataForProperty, config => _.difference(config, sharedValues));
+
+		var maxLength = _(configSpecificValues)
+			.values()
+			.map(value => value.length)
+			.max()
+			.value();
```
teobugslayer: max() is not chainable, therefore .value() should not be needed. Of course, I did not test it :)

lib/project.ts
```diff
@@ -488,6 +491,62 @@ export class Project implements Project.IProject {
...
+	private getConfigurationSpecificDataForProperty(normalizedPropertyName: string): any[] {
+		var configsDataForProperty: any[] = [];
+		var numberOfConfigs = _.keys(this.configurationSpecificData).length;
+
+		_.forEach(_.keys(this.configurationSpecificData), config => {
+			var data = this.configurationSpecificData[config][normalizedPropertyName];
+			configsDataForProperty.push(_.flatten([data]));
+		});
+
+		var sharedValues: string[] = _.intersection.apply(null, configsDataForProperty);
```
Fatme: great! :100:

lib/project.ts
```diff
@@ -488,6 +491,62 @@ export class Project implements Project.IProject {
...
+			.flatten()
+			.value();
+		return _.some(properties, prop => prop === normalizedPropertyName);
+	}
+
+	private getConfigurationSpecificDataForProperty(normalizedPropertyName: string): any[] {
+		var configsDataForProperty: any[] = [];
+		var numberOfConfigs = _.keys(this.configurationSpecificData).length;
+
+		_.forEach(_.keys(this.configurationSpecificData), config => {
```
teobugslayer: .values() maybe?

lib/project.ts
rosen-vladimirov: not working :(

teobugslayer:
:+1:

tytodorov:
:+1:

PR #802, opened 2015-02-10
Fix starting of nativescript projects on device

rosen-vladimirov:
NativeScript projects are successfully installed on emulators and devices, but they are not started. The reason is incorrect startPackageActivity - it is project dependant. Add startPackageActivity for each projectType and set staticConfig's START_PACKAGE_ACTIVITY to it when emulate or deploy command is executed.

Fixes http://teampulse.telerik.com/view#item/285489

### DIFF
--- a/lib/project/web-site-project.ts
+++ b/lib/project/web-site-project.ts
@@ -52,6 +52,10 @@ export class MobileWebSiteProject extends frameworkProjectBaseLib.FrameworkProje
 		return [];
 	}
 
+	public get startPackageActivity(): string {
+		throw new Error("Not applicable.");
+	}
+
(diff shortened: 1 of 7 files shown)


lib/project/web-site-project.ts
```diff
@@ -52,6 +52,10 @@ export class MobileWebSiteProject extends frameworkProjectBaseLib.FrameworkProje
 		return [];
 	}
 
+	public get startPackageActivity(): string {
+		throw new Error("Not applicable.");
```
Fatme: `this.$errors.fail`

Fatme:
:+1:

tytodorov:
:+1:

PR #804, opened 2015-02-10
Help for timeout option passed to emulate and remote commands

Fatme:
http://teampulse.telerik.com/view#item/284307

### DIFF
--- a/resources/help.txt
+++ b/resources/help.txt
@@ -735,7 +735,7 @@ Options:
 
 --[emulate|ios]--
 Usage:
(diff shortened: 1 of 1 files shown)


tytodorov:
:+1:

rosen-vladimirov:
:+1:

rosen-vladimirov:
Do we have teampulse story for this, just to be sure it will be documented.

Fatme:
Yes, we have http://teampulse.telerik.com/view#item/284307

PR #805, opened 2015-02-10
Fix exception on double command arguments

teobugslayer:
Fixes http://teampulse.telerik.com/view#item/280191


rosen-vladimirov:
:+1:

PR #809, opened 2015-02-11
Fix case-sensitivity error when creating project from template

Mitko-Kerezov:
Fixes http://teampulse.telerik.com/view#item/285465
Related to https://github.com/telerik/mobile-cli-lib/pull/187


teobugslayer:
:+1:

PR #810, opened 2015-02-11
Bail out with error message when a parameter on the command line is give...

teobugslayer:
...n more than once.

Add mock for analyticsService to unit tests, so that actual errors could be printed
When an error occures too early, we hit a code which calls .wait() outside of a fibre. Fix it.

Fixes http://teampulse.telerik.com/view#item/280191


rosen-vladimirov:
run ci

lib/bootstrap.ts
```diff
@@ -1,9 +1,9 @@
+require("./common/bootstrap"); // note: here the order is important!
+$injector.require("staticConfig", "./config");
+$injector.require("config", "./config");
 require("./options");
```
rosen-vladimirov: I think you should move userSettingsService before requiring options. When we require options we resolve errors, which resolves analyticsService, which is trying to resolve userSettingsService
```Typescript
$injector.require("config", "./config");
$injector.require("userSettingsService", "./services/user-settings-service");
require("./options");
```
In case you do not do this, you'll receive an error:
Error while reporting exception: Error: unable to resolve userSettingsService (I've executed $ appbuilder emulate android --gosho)

rosen-vladimirov:
:+1:

teobugslayer:
run ci

PR #812, opened 2015-02-12
Preuninstall script should use correct adb path

rosen-vladimirov:
Preuninstall script tries to execute adb from resource/platform-tools... but we have moved the adb file in lib/common/platform-tools... so the script should check the path to adb.


Fatme:
:+1:

tytodorov:
:+1:

teobugslayer:
Should we make a corresponding TNS PR?

PR #813, opened 2015-02-12
Forbid ports below 1024 with sudo

rosen-vladimirov:
If you execute $ sudo appbuilder remote 444 we do not fail. Fix this.

Fixes: http://teampulse.telerik.com/view#item/285912

### DIFF
--- a/lib/commands/remote.ts
+++ b/lib/commands/remote.ts
@@ -55,7 +55,7 @@ export class RemoteCommand implements ICommand {
 				this.$errors.fail("You must specify a valid port number. Valid values are between 1 and 65535.");
 			}
 
-			if (!hostInfo.isWindows() && (parsedPortNumber < 1024) && process.getuid() !== 0) {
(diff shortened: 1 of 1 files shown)


teobugslayer:
:+1:

rosen-vladimirov:
:+1:  from @tytodorov

lib/commands/remote.ts
```diff
@@ -55,7 +55,7 @@ export class RemoteCommand implements ICommand {
 				this.$errors.fail("You must specify a valid port number. Valid values are between 1 and 65535.");
 			}
 
-			if (!hostInfo.isWindows() && (parsedPortNumber < 1024) && process.getuid() !== 0) {
+			if (!hostInfo.isWindows() && (parsedPortNumber < 1024)) {
```
ligaz: Should it or equal to 1024? ... or we have to change the error message :smile:

tytodorov:
:+1:

PR #817, opened 2015-02-13
Update configs according to latest changes

rosen-vladimirov:
- Change local to point to sit-platform. That's why I've removed sit config.
 - Remove dev1 and dev2 as they do not seem to work (also we've used them only once).
 - Remove uat as it does not exist anymore.
 - Change test to point to testtap.telerik.com (old address is redirected to this one).
 - Change dev to point to mocktap.telerik.com NOTE: AppBuilder's deployment configuration on mocktap will be from release. This is local live environment.
 - Add dev3 configuration - NOTE: AppBuilder's deployment configuration is Test.

### DIFF
--- a/config/config-dev.json
+++ b/config/config-dev.json
@@ -1,5 +1,5 @@
 {
-	"AB_SERVER": "icedev.icenium.com",
+	"AB_SERVER": "mocktap.telerik.com",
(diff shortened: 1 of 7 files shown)


teobugslayer:
:+1:

Fatme:
:+1:

PR #826, opened 2015-02-16
Refactor code that shows qr codes to work with lodash 3.2

Fatme:
This code is not working with lodash 3.2. I added `.value()` call to trigger each callback execution.


Fatme:
I am going to fix all places where we have similar code in another PR.

teobugslayer:
Good catch, :+1:

Fatme:
run ci

PR #827, opened 2015-02-16
Adjust infrastructure for Linux BPC

teobugslayer:
Adjust gruntfile for Linux build
build artifacts are copies to R:, and then deployed to Artifactory
Rename local to sit config


rosen-vladimirov:
:+1:

PR #829, opened 2015-02-16
Fix code to work with lodash 3.2

Fatme:


lib/project.ts
```diff
@@ -527,11 +527,10 @@ export class Project implements Project.IProject {
 		var valuesInAllConfigs = _.map(sharedValues, value => helpers.fill(value, numberOfConfigs));
 		var configSpecificValues = _.map(configsDataForProperty, config => _.difference(config, sharedValues));
 
-		var maxLength = _(configSpecificValues)
+		var maxLength = _.chain(configSpecificValues)
```
rosen-vladimirov: why do we need chain ?

teobugslayer:
:+1: after addressing @rosen-vladimirov 's comment

PR #832, opened 2015-02-17
Do not show "null" message on Linux

rosen-vladimirov:
When you do not have connected devices and you execute $appbuilder device on Linux, we print:
null
Cannot find connected ...

Remove printing of error in case there's no error.


teobugslayer:
Please do not forget to create a PR in the TNS project.

Fatme:
:+1:

rosen-vladimirov:
I'll create PR in TNS project after we merge this change in our master branch :)

PR #834, opened 2015-02-18
Don't use global $injector in server

Fatme:


teobugslayer:
:+1: after green build

Fatme:
run ci

PR #839, opened 2015-02-23
Allow projectVersion to be string

rosen-vladimirov:
In some of our samples projectVersion is set to "1". Our code expects number, so the validation fails. Other clients (VSE, Graphite) allow using string for project version, so we use toString in order to allow "1". Json schemas allow both integer and string for projectVersion.

Fixes http://teampulse.telerik.com/view#item/286619


teobugslayer:
:+1:

PR #842, opened 2015-02-23
Show context specific help

rosen-vladimirov:
Refactor several commands in order to be mobile platform specific - deploy, build, livesync, appmanager upload.
Update help.txt to support boolean conditions in order to understand if part of the help.txt should be included in the output.
Use lodash _.template in order to allow using of javascript code directly in help.txt file.
In case you want to call some method that returns string and you need the string in help.txt just write:
Some text... <%= this.$injector.resolve("project").myMethod() %> or even simpler:
Some text... <%= #{project.myMethod} %>
The code between <%= and %> will be executed (it's pure javascript, but you are able to use our $injector and $dynamicHelpService) and the returned string will be placed in the help file.

In case you want to limit some text only for specific case, just write:
<% if (this.$dynamicHelpService.isPlatform("linux")) { %>
In this version of the Telerik AppBuilder CLI, you cannot build and deploy to iOS connected devices on Linux systems.
<% } %>
This way the code in if statement will be shown only in case command is run on linux.

Add localVariables that allow easy call for default scenarios (isLinux, isMacOS, is Windows, isCordova, isNativeScript, isMobileWebsite). In order to use it, just call:
\<% if (isCordova) %\>

Add replace of all entries, which match our dynamicCallRegex with this.$injector.dynamicCall($1).wait()
This way you can call dynamic methods, by just using:
\<%= #{cordovaProject.projectTemplatesString} %\>

Add dynamicHelpProvider.

Add unit tests.

http://teampulse.telerik.com/view#item/283188

### DIFF
--- a/lib/commands/appmanager.ts
+++ b/lib/commands/appmanager.ts
@@ -2,69 +2,36 @@
 "use strict";
 
 import MobileHelper = require("../common/mobile/mobile-helper");
-import constants = require("../common/mobile/constants");
-import util = require("util");
-import options = require("../options");
 
-class AppManagerUploadCommand implements ICommand {
-	constructor(
-		private $config: IConfiguration,
-		private $server: Server.IServer,
-		private $errors: IErrors,
-		private $logger: ILogger,
-		private $project: Project.IProject,
-		private $loginManager: ILoginManager,
-		private $opener: IOpener,
-		private $buildService: Project.IBuildService) {}
+class AppManagerUploadAndroidCommand implements ICommand {
+	constructor(private $appManagerService: IAppManagerService) { }
 
 	execute(args: string[]): IFuture<void> {
-		return (() => {
-			var platform = MobileHelper.validatePlatformName(args[0], this.$errors);
-			this.$project.ensureProject();
-			this.$loginManager.ensureLoggedIn().wait();
-
-			this.$logger.info("Accessing Telerik AppManager store.");
-			this.$server.tam.verifyStoreCreated().wait();
-
-			this.$logger.info("Building release package.");
-			var buildResult = this.$buildService.build({
-				platform: platform,
-				configuration: "Release",
-				provisionTypes: [constants.ProvisionType.Development, constants.ProvisionType.Enterprise, constants.ProvisionType.AdHoc],
-				showWp8SigningMessage: false,
-				buildForTAM: true,
-				downloadFiles: options.download
-			}).wait();
-
-			if (!buildResult[0] || !buildResult[0].solutionPath) {
-				this.$errors.fail({formatStr: "Build failed.", suppressCommandHelp: true});
-			}
-
-			this.$logger.info("Uploading package to Telerik AppManager.");
-			var projectName = this.$project.projectData.ProjectName;
-			var solutionPath = buildResult[0].solutionPath;
-			var projectPath = solutionPath.substr(solutionPath.indexOf("/") + 1);
-			this.$server.tam.uploadApplication(projectName, projectName, projectPath).wait();
-
-			this.$logger.info("Successfully uploaded package.");
-
-			var tamUrl = util.format("%s://%s/appbuilder/Services/tam", this.$config.AB_SERVER_PROTO, this.$config.AB_SERVER);
-			this.$logger.info("Go to %s to manage your apps.", tamUrl);
-			this.$opener.open(tamUrl);
-		}).future<void>()();
+		return this.$appManagerService.upload(MobileHelper.DevicePlatforms[MobileHelper.DevicePlatforms.Android].toLowerCase());
 	}
 
-	allowedParameters : ICommandParameter[] = [new PlatformNameCommandParameter(this.$errors)];
+	allowedParameters: ICommandParameter[] = [];
 }
-$injector.registerCommand("appmanager|upload", AppManagerUploadCommand);
+$injector.registerCommand("appmanager|upload|android", AppManagerUploadAndroidCommand);
 
-class PlatformNameCommandParameter implements ICommandParameter {
-	constructor(private $errors: IErrors) { }
-	mandatory = true;
-	validate(value: string, errorMessage?: string): IFuture<boolean> {
-		return ((): boolean => {
-			MobileHelper.validatePlatformName(value, this.$errors);
-			return true;
-		}).future<boolean>()();
+class AppManagerUploadIosCommand implements ICommand {
+	constructor(private $appManagerService: IAppManagerService) { }
+
+	execute(args: string[]): IFuture<void> {
+		return this.$appManagerService.upload(MobileHelper.DevicePlatforms[MobileHelper.DevicePlatforms.iOS]);
 	}
+
+	allowedParameters: ICommandParameter[] = [];
 }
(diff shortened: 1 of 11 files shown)


lib/services/livesync-service.ts
```diff
@@ -0,0 +1,177 @@
...
+
+import util = require("util");
+import path = require("path");
+var options: any = require("../common/options");
+var gaze = require("gaze");
+import helpers = require("./../helpers");
+import MobileHelper = require("./../common/mobile/mobile-helper");
+import AppIdentifier = require("../common/mobile/app-identifier");
+import constants = require("../common/mobile/constants");
+import commandParams = require("../common/command-params");
```
rosen-vladimirov: commandParams is not used anywhere, remove it

teobugslayer:
Github was hungry and has eaten some characters from the code snippets in the PR description :) Teach it to behave!

lib/commands/deploy.ts
```diff
@@ -8,44 +8,40 @@ import Future = require("fibers/future");
...
 				this.$errors.fail("You will be able to deploy %s based applications in a future release of the Telerik AppBuilder CLI.", this.$project.projectData.Framework);
 			}
-			if (args[0] && !MobileHelper.isPlatformSupported(args[0])) {
-				this.$errors.fail("On your current OS, you cannot deploy apps on connected %s devices.", MobileHelper.normalizePlatformName(args[0]));
+			if(platform && !MobileHelper.isPlatformSupported(platform)) {
+				this.$errors.fail("On your current OS, you cannot deploy apps on connected %s devices.", MobileHelper.normalizePlatformName(platform));
 			}
 
-			this.deployCore(args).wait();
+			this.deployCore(platform).wait();
```
teobugslayer: You can remove the fiber wrappings and return deployCore without waiting it.

lib/commands/deploy.ts
```diff
@@ -8,44 +8,40 @@ import Future = require("fibers/future");
...
 	}
 
-	private deployCore(args: string[]): IFuture<void> {
+	private deployCore(platform: string): IFuture<void> {
 		return ((): void => {
 			if (options.companion) {
 				this.$logger.warn("No deployment necessary when using AppBuilder companion." +
 					" Use the `livesync` command instead to avoid this warning.");
-				this.$commandsService.executeCommandUnchecked("livesync", args).wait();
+				this.$liveSyncService.livesync(platform).wait();
```
teobugslayer: We should remove either the warning or the command. Thoughts?

lib/commands/appmanager.ts
```diff
@@ -1,70 +1,35 @@
...
-			this.$logger.info("Successfully uploaded package.");
+class AppManagerUploadIosCommand implements ICommand {
+	constructor(private $appManagerService: IAppManagerService) { }
 
-			var tamUrl = util.format("%s://%s/appbuilder/Services/tam", this.$config.AB_SERVER_PROTO, this.$config.AB_SERVER);
-			this.$logger.info("Go to %s to manage your apps.", tamUrl);
-			this.$opener.open(tamUrl);
-		}).future<void>()();
+	execute(args: string[]): IFuture<void> {
+		return this.$appManagerService.upload("ios");
```
Fatme: Instead of hardcoded string you can use `MobileHelper.DevicePlatforms.iOS`

lib/services/livesync-service.ts
```diff
@@ -0,0 +1,177 @@
...
+interface IProjectFileInfo {
+	fileName: string;
+	onDeviceName: string;
+	shouldIncludeFile: boolean;
+}
+
+export class LiveSyncService implements ILiveSyncService {
+	private excludedProjectDirsAndFiles = [
+		"app_resources"
+		, "plugins"
```
Fatme: Can we move `,` at the end of the line?

lib/commands/deploy.ts
rosen-vladimirov: Our current code is working this way. The idea is to tell the customers that ther's a difference between deploy and livesync commands, but still do the correct work when they call deploy --companion.
I think we should keep it this way.

lib/commands/appmanager.ts
rosen-vladimirov: I would love to, but MobileHelper.DevicePlatforms is enum and if I try to pass it as string, it returns "0", "1", "2" instead of ios, android and wp8. So I have to apply additional conversions for something simple.

rosen-vladimirov:
github was spanked and now it shows the info correctly. I'll write wiki article how to use the dynamicCalls in our help.txt when we have some more time :)

teobugslayer:
I am not entirely sure that developing our own homegrown templating engine is the best solution. Have you considered using one of the lighter widely used templating engines (http://garann.github.io/template-chooser/) ?

rosen-vladimirov:
run ci

lib/services/livesync-service.ts
```diff
@@ -0,0 +1,176 @@
...
+				this.$errors.fail("You will be able to LiveSync %s based applications to the Companion app in a future release of the Telerik AppBuilder CLI.", this.$project.projectData.Framework);
+			}
+
+			var projectDir = this.$project.getProjectDir().wait();
+
+			var appIdentifier = AppIdentifier.createAppIdentifier(platform,
+				this.$project.projectData.AppIdentifier, options.companion);
+
+			if(options.file) {
+				var isExistFile = this.$fs.exists(options.file).wait();
```
rosen-vladimirov: do not use fs.exists

rosen-vladimirov:
ping @teobugslayer , @Fatme , @ikoevska  - check only the latest commit - "Add microtemplate service" - I've used template engine in it and the whole logic is changed.

resources/help.txt
```diff
@@ -989,28 +1179,44 @@ This command is not applicable to Mobile Website projects.
...
+    --download - If set, downloads the application package to the root of the project.
+--[/]--
+
+--[appmanager|upload|ios]--
+Usage:
+    $ appbuilder appmanager upload ios [--certificate <Certificate ID>] [--provision <Provision ID>] [--download]
+
+Builds the project and uploads the application to Telerik AppManager.
+After the upload completes, you need to go to your app in Telerik AppManager, manually configure it for distribution and publish it.
+<% if (this.$dynamicHelpService.isProjectType(mobileWebsite)) { %>
```
rosen-vladimirov: "mobileWebsite" should be passed as string

rosen-vladimirov:
run ci

lib/dynamic-help-provider.ts
```diff
@@ -0,0 +1,29 @@
...
+			}
+
+			return true;
+		}).future<boolean>()();
+	}
+
+	public getLocalVariables(): IFuture<IDictionary<any>> {
+		return ((): IDictionary<any> => {
+			var localVariables:IDictionary<any> = {};
+			localVariables["isMobileWebsite"] = this.isProjectType(["mobileWebsite"]).wait();
```
Fatme: You can use `projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.MobileWebSite`

teobugslayer:
Excellent work, and the tests are shiny!

:+1: after removing that fs.exists call.

PR #844, opened 2015-02-24
Respect appIdentifier from info.plist file when iOSDeploymentValidator is executed

Fatme:
Fixes http://teampulse.telerik.com/view#item/286588

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -79,6 +79,22 @@ export class Project implements Project.IProject {
 		}).future<string[]>()();
 	}
 
+	public getConfigFileContent(template: string): IFuture<any> {
+		return (() => {
+			var configFile = _.find(this.projectConfigFiles, configFile => configFile.template === template);
+			if(configFile) {
+				try {
+					var configFileContent = this.$fs.readText(configFile.filepath).wait();
+					return configFileContent;
+				} catch(e) {
+					return null;
+				}
+			}
+
+			return null;
+		}).future<any>()();
+	}
+
 	public configurationFilesString(): string {
 		if(!this.frameworkProject) {
 			var result: string[] = [];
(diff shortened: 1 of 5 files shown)


lib/project/nativescript-project.ts
```diff
@@ -55,8 +55,8 @@ export class NativeScriptProject extends frameworkProjectBaseLib.FrameworkProjec
 	public get configFiles():  Project.IConfigurationFile[] {
 		var allConfigFiles = this.$projectFilesManager.availableConfigFiles;
 		return [
-			allConfigFiles["ios-info"],
-			allConfigFiles["android-manifest"]
+			allConfigFiles[this.$projectConstants.IOS_CONFIGURARION_TEMPLATE_NAME],
```
rosen-vladimirov: IOS_CONFIGURATION_TEMPLATE_NAME (configuraRion is not correct IMO)

lib/project/project-constants.ts
```diff
@@ -9,6 +9,10 @@ export class ProjectConstants implements Project.IProjectConstants {
 	public RELEASE_PROJECT_FILE_NAME = ".release.abproject";
 	public CORE_PLUGINS_PROPERTY_NAME = "CorePlugins";
 	public CORDOVA_PLUGIN_VARIABLES_PROPERTY_NAME = "CordovaPluginVariables";
+
+	public IOS_CONFIGURARION_TEMPLATE_NAME = "ios-info";
```
rosen-vladimirov: IOS_CONFIGURATION_TEMPLATE_NAME

lib/project.ts
```diff
@@ -79,6 +79,18 @@ export class Project implements Project.IProject {
 		}).future<string[]>()();
 	}
 
+	public getConfigFileContent(template: string): IFuture<any> {
+		return (() => {
+			var configFile = _.find(this.projectConfigFiles, configFile => configFile.template === template);
+			if(configFile && this.$fs.exists(configFile.filepath).wait()) {
```
teobugslayer: it is a bad code practice to use exists and we should remove it from our code base. It is deprecated in both node 0.12 and io.js. See http://nodejs.org/api/fs.html#fs_fs_exists_path_callback .
Better code would be:

```
try {
  var configFileContent = this.$fs.readText(configFile.filepath).wait();
  return configFileContent;
} catch {
  return null;
}
```

teobugslayer:
:+1:

rosen-vladimirov:
:+1:

ndinev:
:+1:

Fatme:
run ci

PR #845, opened 2015-02-24
Fix edit-configuration command

rosen-vladimirov:
Mobile.Android.ManifestXml.zip on the server had been renamed to Mobile.Cordova.Android.ManifestXml.zip  Also there's additional file - Mobile.NativeScript.Android.ManifestXml.zip  which is used now. This breaks our unit tests and $ appbuilder edit-configuration android-manifest calls.

Add cordova-android-manifest and nativescript-android-manifest configs in order to fix the issue.


Fatme:
:+1:

ndinev:
:+1:

PR #850, opened 2015-02-27
Fix nativescript project version in tests

rosen-vladimirov:
Use correct nativescript version.
I'll fix these tests in master branch to work correctly instead of fixing them before each release.


teobugslayer:
:+1:

tytodorov:
img_20150227_120857

PR #851, opened 2015-02-27
Fix Bash passing unescaped arguments to node

Mitko-Kerezov:
Related to http://teampulse.telerik.com/view#item/287068


teobugslayer:
epic! :+1:

rosen-vladimirov:
:+1:

PR #854, opened 2015-03-05
Respect configuration passed to build method

rosen-vladimirov:
Build method has BuildSettings passed as parameter. Respect the configuration passed in these settings (if it has such). This fixes the problem that calling $ appbuilder appstore upload <app_name> builds the app in Debug configuration and it is rejected by iTunes. The command correctly sets BuildSettings configuration to release, but build method does not respect it.

Fixes http://teampulse.telerik.com/view#item/287256


endarova:
:+1:

PR #856, opened 2015-03-05
Fix gruntfile

rosen-vladimirov:
Support scenario where ResourceDownloadEnvironment and DeploymentEnvironment are different. This is required when testing new release.

Fix default environment to be sit instead of local.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -17,10 +17,12 @@ module.exports = function(grunt) {
 	// To circumvent the issue, hack our environment's PATH and let the OS deal with it, which in practice works
 	process.env.path = process.env.path + (os.platform() === "win32" ? ";" : ":") + "node_modules/.bin";
 
+	var defaultEnvironment = "sit";
+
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -18,9 +18,9 @@ module.exports = function(grunt) {
 	process.env.path = process.env.path + (os.platform() === "win32" ? ";" : ":") + "node_modules/.bin";
 
 	grunt.initConfig({
-		deploymentEnvironment: process.env["DeploymentEnvironment"] || "local",
-		resourceDownloadEnvironment: process.env["ResourceDownloadEnvironment"] || "local",
-		jobName: process.env["JOB_NAME"] || "local",
+		deploymentEnvironment: process.env["DeploymentEnvironment"] || "sit",
```
ligaz: You can extract the default value `"sit"` to a local variable.

rosen-vladimirov:
run ci

ligaz:
:+1:

PR #859, opened 2015-03-06
Use AppIdentifier when setting WP8PackageIdentityName

rosen-vladimirov:
As we've decided to use same default value for WP8PackageIdentityName, which is 1234Telerik.\<appid\>, I've changed the getCorrectWP8PackageIdentityName to use AppIdentifier property. In case the value is longer than 50 symbols, only the first 50 are taken.
Add unit tests.

http://teampulse.telerik.com/view#item/287429


rosen-vladimirov:
run ci

teobugslayer:
:+1:  and I love the tests!

tytodorov:
pr_ceco

PR #861, opened 2015-03-09
Pass correct parameters to simulator to enable feature tracking

rosen-vladimirov:
Simulator has two new parameters that have to be passed: New command line arguments added:
"isfeaturetrackingenabled" : boolean;
"analyticsaccountcode" : string.

Update common lib where a new method to get if analytics is enabled had been added.

http://teampulse.telerik.com/view#item/286925


teobugslayer:
:+1:

PR #862, opened 2015-03-09
Fix hidden mobile website sample

teobugslayer:
1. Get information from all pages when reading data from GitHub
2. changed regex to accommodate for the mobile website template
3. fixed unzipping on node 0.12 under windows (a common lib change)
4. small speedups
5. updated byline to remove warning on install

### DIFF
--- a/lib/services/samples-service.ts
+++ b/lib/services/samples-service.ts
@@ -1,7 +1,6 @@
 ///<reference path="../.d.ts"/>
 "use strict";
 
-import Future = require("fibers/future");
 import path = require("path");
 import util = require("util");
 import os = require("os");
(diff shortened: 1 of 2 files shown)


lib/services/samples-service.ts
```diff
@@ -160,15 +160,32 @@ export class SamplesService implements ISamplesService {
...
+	private getRepositories(gitHubEndpointUrl: string, filter: (repo: any) => boolean): IFuture<string[]> {
+		return (() => {
+			var repos: string[] = [];
+
+			for (var page = 1; ; ++page) {
+				var pagedResult = this.getPagedResult(gitHubEndpointUrl, page).wait();
+				if(_.isEmpty(pagedResult)) {
+					break;
+				}
+				Array.prototype.push.apply(repos, pagedResult);
```
Fatme: Can we use - `repos.push(pagedResult)`?

Fatme:
:+1:  after checking my minor comment

lib/services/samples-service.ts
teobugslayer:
```
[].push([1]) => [ [ 1 ] ]

Array.prototype.push.apply([], [1]) => [ 1 ]
```

We can, bun then we have to _.flatten it.

Fatme: I am voting to use `push` + `_.flatten`. I think this is more readable, but I don't insist.

rosen-vladimirov: I also think push + _.flatten is easier to read, but maybe it will require casting (as we know _.flatten has the great behavior of losing the types.
So I vote :+1: for the current code.

PR #864, opened 2015-03-10
Use new NativeScript template names

rosen-vladimirov:
Use new NativeScript templates. Replace TypeScript written by user with TS which is the actual name of the template. Replace TS. with TypeScript when showing available templates.


Fatme:
:+1:

rosen-vladimirov:
run ci

tytodorov:
:+1:

PR #869, opened 2015-03-12
Replace  appbuilder strings with CLIENT_NAME

rosen-vladimirov:
Common lib is used by several CLIs - appbuilder-cli, nativescript-cli, etc. 

- There are several places where appbuilder is used, while it should be staticConfig.CLIENT_NAME. Fix all of these.
- Replace MobileHelper's exported functions in a new class - MobileHelper. This way we can resolve it from injector and it can easily use staticConfig's CLIENT_NAME. 
- Remove DevicePlatforms enum and add DevicePlatformsConstants class which can be used instead of the enum. 
- Add SYS_REQUIREMENTS_LINK to StaticConfig in order to have different links for each CLI. 
- Add getClientName and getPrivacyPolicyLink methods to IAnalyticsSettingsService.

http://teampulse.telerik.com/view#item/287728

### DIFF
--- a/lib/mobile-platforms-capabilities.ts
+++ b/lib/mobile-platforms-capabilities.ts
@@ -1,0 +1,42 @@
+///<reference path=".d.ts"/>
+"use strict";
+
+export class MobilePlatformsCapabilities implements Mobile.IPlatformsCapabilities {
+	private platformCapabilities: IDictionary<Mobile.IPlatformCapabilities>;
+
+	constructor(private $errors: IErrors) { }
+
+	public getPlatformNames(): string[]{
+		return _.keys(this.getAllCapabilities());
+	}
+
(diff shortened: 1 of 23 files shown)


lib/mobile-platforms-capabilities.ts
```diff
@@ -0,0 +1,53 @@
...
+				cableDeploy: false,
+				companion: true,
+				hostPlatformsForDeploy: ["win32"]
+			}
+		}
+
+		return platformCapabilities;
+	}
+
+	public getPlatformCapabilities(platform: string): Mobile.IPlatformCapabilities {
```
rosen-vladimirov: This method should be part of MobileHelper, there's nothing specific here that requires to have it inside MobilePlatformsCapabilities

teobugslayer:
:+1:

PR #872, opened 2015-03-12
Forbid remote command on Windows and Linux

rosen-vladimirov:
Remote command can work only on MacOS. Forbid using it on any other platform. Move some duplicate logic from execute method of RemoteCommand to validate method of PortCommandParameter.
Minor fix in mobile-platforms-capabilities where the property used to cache platforms, never received value, so the caching was not really working.

Fixes http://teampulse.telerik.com/view#item/287833


teobugslayer:
:+1:

PR #875, opened 2015-03-13
Allow proxy using for http requests

rosen-vladimirov:
Modify config-base according to latest changes in common lib:
Enable proxy using for http requests

Rename the following config options:
 - FIDDLER_HOSTNAME to PROXY_HOSTNAME
 - PROXY_TO_FIDDLER to USE_PROXY

Add PROXY_PORT option to config with default value 8888. Add ConfigBase class which should be used as a base for CLI specific configs.

http://teampulse.telerik.com/view#item/287894


teobugslayer:
@ikoevska, please keep this in mind for future documentation

teobugslayer:
:+1:

rosen-vladimirov:
I've added team pulse story and added tag to-document for it: http://teampulse.telerik.com/view#item/287894

PR #876, opened 2015-03-16
Fix livesync commands

rosen-vladimirov:
Fix bootstrap, where livesync <platform> commands had been required from incorrect file. Fix livesyncService where the platform variable had been reset, so the passed one was not used.

Fixes http://teampulse.telerik.com/view#item/287904


teobugslayer:
:+1: wow!

PR #878, opened 2015-03-16
Remove hardcoded "Debug" configuration when starting ios emulator

Fatme:
Fixes http://teampulse.telerik.com/view#item/287958


rosen-vladimirov:
:+1: after rebasing

PR #880, opened 2015-03-18
At install, ask the user to allow auto completion

teobugslayer:
Implements http://teampulse.telerik.com/view#item/288230


rosen-vladimirov:
:+1:

PR #883, opened 2015-03-24
ScreenBuilder integration

Fatme:
http://teampulse.telerik.com/view#item/286393
http://teampulse.telerik.com/view#item/286391
http://teampulse.telerik.com/view#item/286392

Should be merged after this https://github.com/telerik/mobile-cli-lib/pull/245


lib/commands/debug.ts
```diff
@@ -57,12 +57,14 @@ export class DebugCommand implements ICommand {
...
-		var isRunning = this.$processInfo.isRunning(this.$debuggerPlatformServices.executableName).wait();
-		if (isRunning) {
-			this.$errors.fail({formatStr: "AppBuilder Debugger is currently running and cannot be updated." + os.EOL +
-			"Close it and run $ appbuilder debug again.", suppressCommandHelp: true});
-		}
+	private ensureDebuggerIsNotRunning(): IFuture<void> {
+		return (() => {
+			var isRunning = this.$processInfo.isRunning(this.$debuggerPlatformServices.executableName).wait();
+			if (isRunning) {
+				this.$errors.fail({formatStr: "AppBuilder Debugger is currently running and cannot be updated." + os.EOL +
```
rosen-vladimirov:
```TypeScript
this.$errors.failWithoutHelp(...)
```

lib/commands/project/create.ts
```diff
@@ -0,0 +1,70 @@
...
+				} catch(e) {
+
+				}
+
+				this.$logger.trace("Installing bower");
+				this.$childProcess.exec("bower install", { cwd: projectDirPath }).wait();
+
+				this.initializeProjectFromExistingFiles(this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.Cordova).wait();
+			} catch(err) {
+			}
```
rosen-vladimirov: Maybe we should log.trace the error?

lib/commands/project/create.ts
```diff
@@ -0,0 +1,70 @@
...
+import Future = require("fibers/future");
+import ProjectCommandBaseLib = require("./project-command-base");
+import path = require("path");
+import options = require("./../../options");
+
+export class CreateCommand extends ProjectCommandBaseLib.ProjectCommandBase {
+	private static GENERATOR_NAME = "generator-kendo-ui-mobile";
+
+	constructor(public $project: Project.IProject,
+				public $errors: IErrors,
```
rosen-vladimirov: space/tabs issue maybe

lib/commands/project/create.ts
```diff
@@ -0,0 +1,70 @@
...
+				var projectDirPath = path.resolve(options.path || ".");
+
+				var cliServicePath = path.join(appScaffoldingPath, "lib/cliService");
+				var Scaffolder = require(cliServicePath);
+				var connector = {
+					generatorsCache: path.join(appScaffoldingPath, "cache"),
+					path: projectDirPath,
+					dependencies: ['generator-kendo-ui-mobile@latest'],
+					connect: function _cnnct(done: any) {
+						done();
```
rosen-vladimirov: what if done is a string?

lib/commands/project/create.ts
```diff
@@ -0,0 +1,70 @@
...
+							this.$logger.trace(err);
+							future.throw(err);
+						} else {
+							future.return();
+						}
+					});
+
+					future.wait();
+				} catch(e) {
+
```
rosen-vladimirov: log trace the error?

lib/commands/project/create.ts
```diff
@@ -0,0 +1,70 @@
...
+				}
+
+				this.$logger.trace("Installing bower");
+				this.$childProcess.exec("bower install", { cwd: projectDirPath }).wait();
+
+				this.initializeProjectFromExistingFiles(this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.Cordova).wait();
```
rosen-vladimirov: why do we want to initialize project from existing files in Create command?

lib/commands/simulate.ts
```diff
@@ -60,12 +60,14 @@ export class SimulateCommand implements ICommand {
...
-		var isRunning = this.$processInfo.isRunning(this.$simulatorPlatformServices.executableName).wait();
-		if (isRunning) {
-			this.$errors.fail({formatStr: "AppBuilder Simulator is currently running and cannot be updated." + os.EOL +
-				"Close it and run $ appbuilder simulate again.", suppressCommandHelp: true});
-		}
+	private ensureSimulatorIsNotRunning(): IFuture<void> {
+		return (() => {
+			var isRunning = this.$processInfo.isRunning(this.$simulatorPlatformServices.executableName).wait();
+			if (isRunning) {
+				this.$errors.fail({formatStr: "AppBuilder Simulator is currently running and cannot be updated." + os.EOL +
```
rosen-vladimirov: failWithoutHelp

lib/commands/project/create.ts
Fatme: done is always callback, it is from appscaffolding api

lib/dependency-config.ts
```diff
@@ -0,0 +1,35 @@
+///<reference path=".d.ts"/>
+"use strict";
+
+import path = require("path");
+
+export class DependencyConfigService implements IDependencyConfigService {
```
rosen-vladimirov: As this is a service, shouldn't we move it to services folder?

lib/commands/project/create.ts
Fatme: Because this files are generated from kendo-ui-generator. We want to create appbuilder project and just init in directory with already created files is enough.

lib/services/generator-extensions-service.ts
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	public getGeneratorCachePath(generatorName: string, appScaffoldingPath: string): string {
+		return path.join(appScaffoldingPath, "latest", "node_modules", generatorName);
+	}
+
+	public prepareGenerator(generatorName: string, appScaffoldingPath: string): IFuture<void> {
+		return (() => {
+			var generatorConfig = this.$dependencyConfigService.getGeneratorConfig(generatorName).wait();
+			generatorConfig.pathToSave = this.getGeneratorCachePath(generatorName, appScaffoldingPath);
```
rosen-vladimirov: if generatorConfig is undefined (getGeneratorConfig  is using _.find which will return undefined in case it cannot find the searched element), this call will fail

lib/dependency-config.ts
Fatme: Yes, we should :smile_cat: I'll move it.

lib/commands/project/create.ts
Fatme: Yes, I've fixed it.

lib/dependency-config.ts
```diff
@@ -0,0 +1,35 @@
...
+
+	public getAppScaffoldingConfig(): IFuture<IAppScaffoldingConfig> {
+		return (() => {
+			var dependencyConfig = this.getDependencyConfigContent().wait();
+			return dependencyConfig.appScaffolding;
+		}).future<IAppScaffoldingConfig>()();
+	}
+
+	private getDependencyConfigContent():IFuture<any> {
+		return this.$fs.readJson(this.dependencyConfigFilePath);
```
rosen-vladimirov: It is a good idea to cache the result, as two different methods are calling ```getDependencyConfigContent```. This way we'll read the file only once :)

lib/services/build.ts
```diff
@@ -34,7 +34,7 @@ export class BuildService implements Project.IBuildService {
 		private $multipartUploadService: IMultipartUploadService,
 		private $jsonSchemaValidator: IJsonSchemaValidator,
 		private $projectConstants: Project.IProjectConstants,
-		private $mobileHelper: Mobile.IMobileHelper) { }
+			private $mobileHelper: Mobile.IMobileHelper) { }
```
rosen-vladimirov: space/tabs

lib/services/generator-extensions-service.ts
Fatme: Yes, you're right. If generatorConfig is undefined, this means that `dependecy-config.json` file is empty, which is unexpected behavior. But as we always expect to have config files I didn't add check for that.

lib/commands/project/create.ts
teobugslayer: Then declare it as Function? Like: ``` _cnnct(done: Function) ```

rosen-vladimirov: :+1:  for @teobugslayer  suggestion

lib/commands/project/create.ts
```diff
@@ -0,0 +1,70 @@
...
+		private $screenBuilderService: IScreenBuilderService) {
+		super($project, $errors, $logger);
+	}
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			try {
+				this.$screenBuilderService.prepareScreenBuilder(CreateCommand.GENERATOR_NAME).wait();
+				var appScaffoldingPath = this.$appScaffoldingExtensionsService.appScaffoldingPath;
+				var projectDirPath = path.resolve(options.path || ".");
```
teobugslayer: Currently we are creating files in the current dir. This will be fixed after this PR with changes on the ScreenBuilder side.

lib/commands/project/create.ts
```diff
@@ -0,0 +1,70 @@
...
+							future.return();
+						}
+					});
+
+					future.wait();
+				} catch(e) {
+					this.$errors.fail("Error while creating kendo project: " + e);
+				}
+
+				this.$logger.trace("Installing bower");
```
teobugslayer: This is not installing bower. This is using bower to install project dependencies.

lib/services/extensions-service-base.ts
```diff
@@ -0,0 +1,97 @@
...
+import options = require("../common/options");
+import helpers = require("../helpers");
+
+export class ExtensionsServiceBase {
+	private extensionVersions: IStringDictionary = {};
+
+	constructor(public $fs: IFileSystem,
+				public $httpClient: Server.IHttpClient,
+				public $logger: ILogger) {
+		if (this.$fs.exists(this.versionsFile).wait()) {
```
rosen-vladimirov: Instead of using fs.exists, you can try reading and catch ENOENT error

lib/commands/project/init-command-base.ts
```diff
@@ -0,0 +1,94 @@
...
+		protected $fs: IFileSystem,
+		protected $logger: ILogger,
+		protected $mobileHelper: Mobile.IMobileHelper) {
+		super($project, $errors, $logger);
+
+		this.projectDir = $project.getNewProjectDir();
+		this.tnsModulesDir = new FileDescriptor(path.join(this.projectDir, "tns_modules"), "directory");
+		this.bootstrapFile = new FileDescriptor(path.join(this.projectDir, "app", "bootstrap.js"), "file");
+		this.indexHtml = new FileDescriptor(path.join(this.projectDir, "index.html"), "file");
+		this.cordovaFiles = _.map(this.$mobileHelper.platformNames, platform => {
```
teobugslayer: you may omit { return ... }

lib/services/extensions-service-base.ts
```diff
@@ -0,0 +1,97 @@
...
+				this.$logger.trace("Extension path for %s: %s", packageName, extensionPath);
+
+				try {
+					this.downloadPackage(extensionData.downloadUri, zipFileName).wait();
+					this.$fs.unzip(zipFileName, extensionPath).wait();
+					this.$fs.deleteFile(zipFileName).wait();
+					this.extensionVersions[packageName] = extensionVersion;
+					this.saveVersionsFile().wait();
+				} catch(err) {
+					this.$fs.deleteDirectory(extensionPath).wait();
```
rosen-vladimirov: maybe we should delete the zipFileName as well

lib/services/screen-builder-service.ts
```diff
@@ -0,0 +1,33 @@
...
+			this.$fs.ensureDirectoryExists(cacheFolderPath).wait();
+			this.$fs.ensureDirectoryExists(cachedGeneratorPath).wait();
+
+			this.$generatorExtensionsService.prepareGenerator(generatorName, cachedGeneratorPath).wait();
+			this.npmInstall(this.$generatorExtensionsService.getGeneratorCachePath(generatorName, cachedGeneratorPath)).wait();
+		}).future<void>()();
+	}
+
+	private npmInstall(currentWorkingDirectory: string): IFuture<void> {
+		return this.$childProcess.exec("npm install", {cwd: currentWorkingDirectory });
```
rosen-vladimirov: will this work in case npm is installed with sudo

lib/dependency-config.ts
teobugslayer: Or use _.once (For a second I thought that there will be a PR without me suggesting using a lodash method somewhere. :sunglasses: )

lib/dependency-config.ts
```diff
@@ -0,0 +1,35 @@
...
+
+	public get dependencyConfigFilePath(): string {
+		return path.join(__dirname, "../config", DependencyConfigService.DEPENDENCY_CONFIG_NAME);
+	}
+
+	public getGeneratorConfig(generatorName: string): IFuture<IGeneratorConfig> {
+		return (() => {
+			var dependencyConfigContent = this.getDependencyConfigContent().wait();
+
+			var generatorConfig = _.find(dependencyConfigContent.generators, (generator: IGeneratorConfig) => generator.name === generatorName);
```
teobugslayer: ```generatorName``` ultimately comes raw from the command line. The user can wRiTe LiKe ThIS. Should we massage the input before comparing it (maybe .toLowerCase())?

lib/services/extensions-service-base.ts
```diff
@@ -0,0 +1,97 @@
...
+
+			if( helpers.versionCompare(cachedVersion, extensionVersion) < 0) {
+				this.$logger.info("Updating %s package...", packageName);
+				var zipFileName = path.join(this.cacheDir, packageName + ".zip");
+
+				if(beforeDownloadAction) {
+					beforeDownloadAction().wait();
+				}
+
+				if(this.$fs.exists(extensionPath).wait()) {
```
teobugslayer: The if is superfluous and can be removed. See [rimraf](https://www.npmjs.com/package/rimraf)

lib/services/extensions-service-base.ts
```diff
@@ -0,0 +1,97 @@
...
+			this.$logger.debug("Server version: %s", extensionVersion);
+
+			if (this.extensionVersions[packageName]) {
+				cachedVersion = this.extensionVersions[packageName];
+				this.$logger.debug("Cached version is: %s", cachedVersion);
+			}
+
+			if( helpers.versionCompare(cachedVersion, extensionVersion) < 0) {
+				this.$logger.info("Updating %s package...", packageName);
+				var zipFileName = path.join(this.cacheDir, packageName + ".zip");
```
teobugslayer: this looks like job for temp()

lib/services/extensions-service-base.ts
```diff
@@ -0,0 +1,97 @@
...
+	private get versionsFile(): string {
+		return path.join(this.cacheDir, "extension-versions.json");
+	}
+
+	private saveVersionsFile() : IFuture<void> {
+		return this.$fs.writeJson(this.versionsFile, this.extensionVersions);
+	}
+
+	private downloadPackage(downloadUri: string, zipFileName: string): IFuture<void> {
+		return ((): void => {
```
teobugslayer: You can remove the return () wrapper

lib/services/generator-extensions-service.ts
teobugslayer: I think the check should be when reading the config file content and the code after that can assume that everything is correct.

lib/dependency-config.ts
Fatme: `generatorName` doesn't come form command line. We have hardcoded it in create command.

lib/services/extensions-service-base.ts
Fatme: We've deleted it after unzipping `this.$fs.deleteFile(zipFileName).wait();`

rosen-vladimirov: But if the unzip fails, we'll not delete it

lib/services/generator-extensions-service.ts
Fatme: I've added checks in `getGeneratorConfig` itself if the passed generator name is correct and also if the `dependency-generator.json` file exists.

lib/services/extensions-service-base.ts
Fatme: I'll implement this in another PR because we've already have too much "ENOENT" checks in our code.

lib/services/extensions-service-base.ts
Fatme: I fixed it.

Fatme:
run ci

lib/commands/dev/prepackage.ts
```diff
@@ -32,6 +32,7 @@ export class PrePackageCommand implements ICommand {
 			//Cordova files have to be downloaded after cordova migration data so we know which cordova versions we support
 			this.$logger.info("Downloading cordova.js files.");
 			this.$resourceDownloader.downloadCordovaJsFiles().wait();
+			this.$logger.info("Preparing screenBuilder...");
```
ligaz: The correct casing for Screen Builder is `Screen Builder`.

teobugslayer:
There is no commandsServiceProvider registered in tests and they currently fail

lib/commands/project/create.ts
```diff
@@ -0,0 +1,36 @@
...
+		super($project, $errors, $logger);
+	}
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			var projectDirPath = path.resolve(options.path || ".");
+			this.$screenBuilderService.prepareAndGeneratePrompt(CreateCommand.GENERATOR_NAME).wait();
+
+			this.$logger.trace("Installing project dependencies using bower");
+			this.$childProcess.exec("bower install", { cwd: projectDirPath }).wait();
```
ligaz: Isn't it better to move this call into `screenBuilderService`? It looks like its responsibility to do the bower mumbo jumbo.

lib/services/extensions-service-base.ts
```diff
@@ -0,0 +1,99 @@
...
+					this.downloadPackage(extensionData.downloadUri, zipFileName).wait();
+					this.$fs.unzip(zipFileName, extensionPath).wait();
+					this.$fs.deleteFile(zipFileName).wait();
+					this.extensionVersions[packageName] = extensionVersion;
+					this.saveVersionsFile().wait();
+				} catch(err) {
+					this.$fs.deleteDirectory(extensionPath).wait();
+					throw err;
+				}
+				this.$logger.info("Finished updating %s package.", packageName);
```
teobugslayer: There's a significant delay after printing this message on the console under linux. I suggest we add some busy indicator to tell the user that we are working.

lib/commands/project/create.ts
Fatme: Actually I also was thinking to move it into `screenBuilderService`, but we need it only when we create application. `prepareAndGeneratePrompt` function is called for example when the user adds view, dataProvider, form, field and etc. In these cases we don't need to install bower.

teobugslayer:
:+1:

lib/providers/commands-service-provider.ts
```diff
@@ -0,0 +1,15 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+class CommandsServiceProvider implements ICommandsServiceProvider {
+	constructor(public $screenBuilderService: IScreenBuilderService) { }
+
+	public allDynamicCommands(): IFuture<string[]> {
```
rosen-vladimirov: in ICommandsServiceProvider the method is getDynamicCommands

lib/services/dependency-config.ts
```diff
@@ -0,0 +1,62 @@
...
+				try {
+					this.dependencyConfigCache = this.$fs.readJson(this.dependencyConfigFilePath).wait();
+				} catch(e) {
+					this.$errors.fail("Unable to process config/dependency-config.json file. Check if it exists and try again.");
+				}
+			}
+
+			return this.dependencyConfigCache;
+		}).future<any>()();
+
```
rosen-vladimirov: a lot of blank lines

lib/services/extensions-service-base.ts
rosen-vladimirov: uhm... where

rosen-vladimirov:
:+1:

PR #884, opened 2015-03-25
Fix hooks

rosen-vladimirov:
When the command is hierarchical, hooks are not executed correctly. Remove everything after the pipe in command name when initializing the hook.


rosen-vladimirov:
I'll write unit/integration tests in a separate PR

Fatme:
:+1:

PR #886, opened 2015-03-26
Remove enoent checks, use tryExecuteFileOperation function instead

Fatme:
Should be merged after this https://github.com/telerik/mobile-cli-lib/pull/244


teobugslayer:
:+1:

PR #893, opened 2015-03-31
Marketplace versioning

Fatme:
http://teampulse.telerik.com/view#item/279691

### DIFF
--- a/lib/commands/plugin/list-plugin.ts
+++ b/lib/commands/plugin/list-plugin.ts
@@ -2,6 +2,7 @@
 "use strict";
 
 import options = require("../../common/options");
+import pluginsDataLib = require("./../../plugins-data");
 
 export class ListPluginCommand implements ICommand {
 	constructor(private $pluginsService: IPluginsService) { }
@@ -9,7 +10,25 @@ export class ListPluginCommand implements ICommand {
 	public execute(args: string[]): IFuture<void> {
 		return (() => {
 			var plugins = options.available ? this.$pluginsService.getAvailablePlugins() : this.$pluginsService.getInstalledPluginsEnabledAtLeastInOneConfiguration();
-			this.$pluginsService.printPlugins(plugins);
+			if(options.available) {
+				// Group marketplace plugins
+				var marketplacePlugins = _.filter(plugins, (pl) => pl.type === pluginsDataLib.PluginType.MarketplacePlugin);
+				var output = _.filter(plugins, pl => pl.type === pluginsDataLib.PluginType.CorePlugin || pl.type === pluginsDataLib.PluginType.AdvancedPlugin);
+
+				var groups = _.groupBy(marketplacePlugins, (plugin: IPlugin) => plugin.data.Identifier);
+				_.each(groups, (group: any) => {
+					var defaultData = _.find(group, (gr: IPlugin) => {
+						var pvd = (<any>gr).pluginVersionsData;
+						return pvd && gr.data.Version === pvd.DefaultVersion;
+					});
+					if(defaultData) {
+						output.push(defaultData);
+					}
+				});
+				this.$pluginsService.printPlugins(output);
+			} else {
+				this.$pluginsService.printPlugins(plugins);
+			}
 		}).future<void>()();
 	}
 
--- a/lib/services/marketplace-plugins-service.ts
+++ b/lib/services/marketplace-plugins-service.ts
@@ -5,27 +5,17 @@ import util = require("util");
 import PluginsDataLib = require("./../plugins-data");
 
 export class MarketplacePluginsService implements ICordovaPluginsService {
-	private static MARKET_PLACE_PLUGINS_URL = "https://plugins.telerik.com/api/plugins";
 
-	constructor(private $httpClient: Server.IHttpClient,
-		private $server: Server.IServer,
+	constructor(private $server: Server.IServer,
 		private $project: Project.IProject,
 		private $projectConstants: Project.IProjectConstants) { }
 
 	public getAvailablePlugins(): IFuture<any> {
-		return (() => {
-			return JSON.parse(this.$httpClient.httpRequest(MarketplacePluginsService.MARKET_PLACE_PLUGINS_URL).wait().body);
-		}).future<any>()();
+		return this.$server.cordova.getMarketplacePluginsData();
 	}
 
-	public createPluginData(plugin: any): IFuture<IMarketplacePlugin> {
-		return (() => {
-			var rowPluginData = this.$server.cordova.getMarketplacePluginData(plugin.uniqueId, plugin.pluginVersion).wait();
-			if(!rowPluginData.Url) {
-				rowPluginData.Url = plugin.repositoryUrl;
-			}
-			return new PluginsDataLib.MarketplacePluginData(rowPluginData, plugin.downloadsCount, plugin.demoAppRepositoryLink, plugin.publisher, this.$project, this.$projectConstants);
-		}).future<IMarketplacePlugin>()();
+	public createPluginData(plugin: Server.MarketplacePluginVersionsData): IMarketplacePlugin[] { // DefaultVersion, Identifier, Versions
+		return _.map(plugin.Versions, (pluginVersionData) => new PluginsDataLib.MarketplacePluginData(plugin, pluginVersionData, this.$project, this.$projectConstants));
 	}
 }
 $injector.register("marketplacePluginsService", MarketplacePluginsService);
--- a/lib/services/plugins-service.ts
+++ b/lib/services/plugins-service.ts
@@ -4,6 +4,7 @@
 import os = require("os");
 import util = require("util");
 import options = require("../common/options");
+import pluginsDataLib = require("../plugins-data");
 
 export class PluginsService implements IPluginsService {
 	private static CORE_PLUGINS_PROPERTY_NAME = "CorePlugins";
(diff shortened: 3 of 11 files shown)


lib/services/marketplace-plugins-service.ts
```diff
@@ -2,30 +2,22 @@
 "use strict";
 
 import util = require("util");
+import options = require("../options");
```
rosen-vladimirov: Do you need these options? As far as I see they are not used anywhere

lib/services/marketplace-plugins-service.ts
```diff
@@ -2,30 +2,22 @@
...
-	public createPluginData(plugin: any): IFuture<IMarketplacePlugin> {
-		return (() => {
-			var rowPluginData = this.$server.cordova.getMarketplacePluginData(plugin.uniqueId, plugin.pluginVersion).wait();
-			if(!rowPluginData.Url) {
-				rowPluginData.Url = plugin.repositoryUrl;
-			}
-			return new PluginsDataLib.MarketplacePluginData(rowPluginData, plugin.downloadsCount, plugin.demoAppRepositoryLink, plugin.publisher, this.$project, this.$projectConstants);
-		}).future<IMarketplacePlugin>()();
+	public createPluginData(plugin: Server.MarketplacePluginVersionsData): IMarketplacePlugin { // DefaultVersion, Identifier, Versions
+		var pluginData = _.find(plugin.Versions, pl => pl.Version === (<any>plugin).DefaultVersion);
```
rosen-vladimirov: maybe you can extract `(<any>plugin).DefaultVersion` out of the `_.find`

lib/services/plugins-service.ts
```diff
@@ -52,7 +52,37 @@ export class PluginsService implements IPluginsService {
...
+			var version = parts.length > 1 ? parts[1] : "";
+
+			if(!version) {
+				var toLowerCasePluginName = pluginName.toLowerCase();
+
+				var versions:any[] = [];
+				_.each(this.getAvailablePlugins(), (plugin:IPlugin) => {
+					if (plugin.data.Name.toLowerCase() === toLowerCasePluginName || plugin.data.Identifier.toLowerCase() === toLowerCasePluginName) {
+						var pluginVersionsData = (<IMarketplacePlugin>plugin).pluginVersionsData;
+						_.each(pluginVersionsData.Versions, p => {
```
rosen-vladimirov: maybe
```TypeScript
versions = _.map(pluginVersionsData.Versions, p => { return {name: p.Version, value: p.Version}; });
```

lib/services/plugins-service.ts
```diff
@@ -52,7 +52,37 @@ export class PluginsService implements IPluginsService {
...
+
+				version = versions[0].value;
+				if (versions.length > 1) {
+					var promptData = [{
+						type: "list",
+						name: "version",
+						message: "Which plugin version would you like to use?",
+						choices: versions
+					}];
+					var answer = this.$prompter.get(promptData).wait();
```
rosen-vladimirov: what is the format of the answer here? Is it possible the return value to be invalid version?

Fatme: No, it is not possible. The user is not able to type version. I list all available versions and the user can select only one from them. - https://github.com/Icenium/icenium-cli/pull/893#discussion-diff-27793900R75

rosen-vladimirov:
I've found some issues after marketplace plugin is added

lib/services/plugins-service.ts
```diff
@@ -52,7 +52,37 @@ export class PluginsService implements IPluginsService {
...
+				var toLowerCasePluginName = pluginName.toLowerCase();
+
+				var versions:any[] = [];
+				_.each(this.getAvailablePlugins(), (plugin:IPlugin) => {
+					if (plugin.data.Name.toLowerCase() === toLowerCasePluginName || plugin.data.Identifier.toLowerCase() === toLowerCasePluginName) {
+						var pluginVersionsData = (<IMarketplacePlugin>plugin).pluginVersionsData;
+						versions = _.map(pluginVersionsData.Versions, p => {
+							return {name: p.Version, value: p.Version};
+						});
+					}
```
teobugslayer: Can we stop executing the .each() when the if succeeds?

lib/services/plugins-service.ts
```diff
@@ -52,7 +52,37 @@ export class PluginsService implements IPluginsService {
...
+
+			if(!version) {
+				var toLowerCasePluginName = pluginName.toLowerCase();
+
+				var versions:any[] = [];
+				_.each(this.getAvailablePlugins(), (plugin:IPlugin) => {
+					if (plugin.data.Name.toLowerCase() === toLowerCasePluginName || plugin.data.Identifier.toLowerCase() === toLowerCasePluginName) {
+						var pluginVersionsData = (<IMarketplacePlugin>plugin).pluginVersionsData;
+						versions = _.map(pluginVersionsData.Versions, p => {
+							return {name: p.Version, value: p.Version};
```
teobugslayer: you can shorten the lambda a bit

rosen-vladimirov:
`$ appbuilder plugin remove com.phonegap.plugins.PushPlugin@2.4.2
Gathering information for plugins...
Plugin com.phonegap.plugins.PushPlugin@2.4.2 is not installed`

I'm not sure we should support this, but in my .debug.abproject I see com.phonegap.plugins.PushPlugin@2.4.2 and I've copied it from there and tried to remove it

rosen-vladimirov:
I've tried the following: 
* create new project, its Cordova version is 3.7.0
* add PushPlugin - version 2.2.1
* migrate my Cordova to 3.2.0
PushPlugin is missing now, I cannot find it in my .debug.abproject and .release.abproject

rosen-vladimirov:
`$ appbuilder plugin --available`
lists each plugin several times, one for each of its versions:

   Plugin: SocialSharing
   Identifier: nl.x-services.plugins.socialsharing
   Version: 4.3.13
   Url: https://github.com/Telerik-Verified-Plugins/SocialSharing
   Downloads count: 1144
   Available versions: 4.3.8, 4.3.6, 4.3.0

   Plugin: SocialSharing
   Identifier: nl.x-services.plugins.socialsharing
   Version: 4.3.8
   Url: https://github.com/Telerik-Verified-Plugins/SocialSharing
   Downloads count: 1144
   Available versions: 4.3.13, 4.3.6, 4.3.0

   Plugin: SocialSharing
   Identifier: nl.x-services.plugins.socialsharing
   Version: 4.3.6
   Url: https://github.com/Telerik-Verified-Plugins/SocialSharing
   Downloads count: 1144
   Available versions: 4.3.13, 4.3.8, 4.3.0

   Plugin: SocialSharing
   Identifier: nl.x-services.plugins.socialsharing
   Version: 4.3.0
   Url: https://github.com/Telerik-Verified-Plugins/SocialSharing
   Downloads count: 1144
   Available versions: 4.3.13, 4.3.8, 4.3.6

lib/services/plugins-service.ts
```diff
@@ -178,32 +216,33 @@ export class PluginsService implements IPluginsService {
 	private createPluginsData(pluginsService: ICordovaPluginsService): IFuture<void> {
 		return (() => {
 			var plugins = pluginsService.getAvailablePlugins().wait();
-
 			_.each(plugins, (plugin: Server.CordovaPluginData) => {
-
 				try {
-					var pluginData = pluginsService.createPluginData(plugin).wait();
-				} catch(e) {
+					var data = pluginsService.createPluginData(plugin);
```
rosen-vladimirov: as this var is used outside of the try-catch, it should be declared before try :)

lib/services/plugins-service.ts
```diff
@@ -52,7 +52,37 @@ export class PluginsService implements IPluginsService {
 				this.$errors.fail("Plugin %s is already installed", pluginName);
 			}
 
-			this.configurePlugin(pluginName).wait();
+			var parts = pluginName.split("@");
+			pluginName = parts[0];
+			var version = parts.length > 1 ? parts[1] : "";
```
teobugslayer: Maybe ```version = parts[1] || ""```?

lib/services/plugins-service.ts
teobugslayer: The code for managing plugin versions and querying the user is in the same function. Can we split them?

lib/services/plugins-service.ts
```diff
@@ -178,32 +216,33 @@ export class PluginsService implements IPluginsService {
...
-					_.each(configurations, (configuration:string) => {
-						var configData = this.$project.configurationSpecificData[configuration];
-						if (configData) {
-							var corePlugins = configData[PluginsService.CORE_PLUGINS_PROPERTY_NAME];
-							if (corePlugins && _.contains(corePlugins, projectDataRecord)) {
-								pluginData.configurations.push(configuration);
+				_.each(data, pluginData => {
+					if (pluginData && pluginData.data) {
+						var projectDataRecord = pluginData.toProjectDataRecord();
+						var configurations = _.keys(this.$project.configurationSpecificData);
```
teobugslayer: You do not need to call _.keys - _.each gives you the keys as well

lib/services/plugins-service.ts
Fatme: JavaScript is powerful and the var can be declared in try catch block and used outside it. :smile_cat: I prefer it because it's shorter.

Fatme:
@rosen-vladimirov,

1. We support adding, removing, configuring plugins by name or identifier, so you can type ` $ appbuilder plugin remove pushplugin` or ` $appbuilder plugin remove com.phonegap.plugin.PushPlugin`. I think that it is unneeded to support this syntax `$ appbuilder plugin remove com.phonegap.plugins.PushPlugin@2.4.2`. This record is just for internal usage.

2. About your test with creating project with latest cordova version and migrating to earlier - we're migrating only cordova plugins, we don't have logic for migrating marketplace plugins. The problem here is that we're overwriting corePlugins property with migrated cordova plugins and this way we're removing all installed marketplace plugins. I'll log a story to implement it.

3. You're correct that `plugin --available` command shows some of the plugins more than once. I've already fixed it.

lib/services/plugins-service.ts
Fatme: Hm, I am not sure that I understand correctly what you mean. I need `configurations` so I use `_.keys(this.$project.configurationSpecificData)`, but the `_.each` enumerates `pluginData`

rosen-vladimirov: Maybe @teobugslayer  was talking about next each, where you are `eaching` (I love this new word :smiley_cat: ) configurations and you are also using 
```TypeScript
var configData = this.$project.configurationSpecificData[configuration];
```
The idea is to take all configurations here (where @teobugslayer  commented), not only the keys, as it looks like you need both keys and values.

rosen-vladimirov:
I'm fine with the changes, :+1:  from me! Great work :smile_cat:

lib/commands/plugin/list-plugin.ts
```diff
@@ -9,7 +10,25 @@ export class ListPluginCommand implements ICommand {
...
 		return (() => {
 			var plugins = options.available ? this.$pluginsService.getAvailablePlugins() : this.$pluginsService.getInstalledPluginsEnabledAtLeastInOneConfiguration();
-			this.$pluginsService.printPlugins(plugins);
+			if(options.available) {
+				// Group marketplace plugins
+				var marketplacePlugins = _.filter(plugins, (pl) => pl.type === pluginsDataLib.PluginType.MarketplacePlugin);
+				var output = _.filter(plugins, pl => pl.type === pluginsDataLib.PluginType.CorePlugin || pl.type === pluginsDataLib.PluginType.AdvancedPlugin);
+
+				var groups = _.groupBy(marketplacePlugins, (plugin: IPlugin) => plugin.data.Identifier);
+				_.each(groups, (group: any) => {
```
teobugslayer: Not a merge stopper, but _.transform or _.reduce are better fits here

lib/services/plugins-service.ts
teobugslayer: This is considered bad coding practice and is explicitly mentioned in our coding guidelines: https://github.com/Icenium/Ice/wiki/JavaScript-Coding-Guidelines#all-variables-should-be-declared-on-the-top-of-the-function-with-only-one-var

teobugslayer:
I am happy with the PR, :+1:

PR #895, opened 2015-03-31
Update typescript declarations file

rosen-vladimirov:
Update the file in our resources which is used for TypeScript projects. Set the return value of setTimeout and setInterval to any in order to support cases when it returns number and Timer. Add setImmediate and clearImmediate methods.


Mitko-Kerezov:
:+1:

PR #897, opened 2015-04-01
Fix installation and uninstall of CLI

rosen-vladimirov:
Fix installation of appbuilder-cli by removing ProjectTemplates, ItemTemplates, etc. directories from .npmignore file. This way our resources are uploaded to npm and we are working fine.
Fix preuninstall script which was failing on several places (missing lodash as it was not required, unable to resolve injector, etc.) In order to fix the issues, new command  "dev-preuninstall" is introduced. This way injector can be resolved.

### DIFF
--- a/preuninstall.js
+++ b/preuninstall.js
@@ -1,24 +1,10 @@
 var child_process = require("child_process");
-var util = require("util");
-var path = require("path");
-var fs = require("fs");
-var options = require("./lib/options");
 
-var adbPath = util.format("resources/platform-tools/android/%s/adb", process.platform);
-var executableAdbPath = process.platform === "win32" ? adbPath + ".exe" : adbPath;
-if(!fs.existsSync(executableAdbPath)) {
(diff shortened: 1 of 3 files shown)


preuninstall.js
```diff
@@ -1,24 +1,9 @@
...
-	adbPath = "lib/common/" + adbPath;
-}
-
-var killAdbServerCommand = util.format("\"%s\" %s", adbPath, " kill-server");
-child_process.exec(killAdbServerCommand, function(error) {
-	if(error) {
-		console.log(error.toString());
+var child = child_process.exec("node bin/appbuilder.js dev-preuninstall", function (error) {
+	if (error) {
+		console.error("Failed to complete all pre-uninstall steps.");
```
teobugslayer: But why? Give me the error! 

Joke aside, please print the error message.

Fatme:
:+1:

PR #900, opened 2015-04-02
Style html generated pages

vladzagorski:
Add common styles for the html generated pages


rosen-vladimirov:
Nice, I really like these styles and we'll definitely have the best help of a command line tool. We've checked them with @Fatme  and we have some questions (ideas for improvement):
* Can we add AppBuilder logo in basic_page.html, so on each page we'll have it at the top of the content?
* Can we add favicon, maybe Telerik's or AppBuilder's log?
* Can we limit the pages to take less space, currently they are expanded to the whole available space, but we suggest to narrow it down, for example to 80% maybe? 

What do you think about all of these :)

rosen-vladimirov:
@ikoevska , I agree to use H3s for Related Articles instead of H4, this way we don't have to change H4's styles.

vladzagorski:
@rosen-vladimirov, @ikoevska - will add the proposed changes. About the command syntax, I think it should be more visible, similar to the npm documentation but we can use the look from the docs for consistency.

rosen-vladimirov:
:+1:

PR #902, opened 2015-04-02
Fix wrong android sdk detection on mac - sync with latest common

teobugslayer:
Fixes http://teampulse.telerik.com/view#item/289312


rosen-vladimirov:
:+1:

PR #906, opened 2015-04-07
Fix broken integration tests

Fatme:
The templates were changed on the server and we need to update our test templates.


rosen-vladimirov:
:+1:

PR #909, opened 2015-04-08
Use latest server api

Fatme:
http://teampulse.telerik.com/view#item/289686

### DIFF
--- a/lib/services/marketplace-plugins-service.ts
+++ b/lib/services/marketplace-plugins-service.ts
@@ -11,7 +11,7 @@ export class MarketplacePluginsService implements ICordovaPluginsService {
 		private $projectConstants: Project.IProjectConstants) { }
 
 	public getAvailablePlugins(): IFuture<any> {
-		return this.$server.cordova.getMarketplacePluginsData();
+		return this.$server.cordova.getMarketplacePluginsData(this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.Cordova);
 	}
 
 	public createPluginData(plugin: Server.MarketplacePluginVersionsData): IMarketplacePlugin[] { // DefaultVersion, Identifier, Versions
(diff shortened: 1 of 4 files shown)


lib/services/marketplace-plugins-service.ts
```diff
@@ -11,7 +11,7 @@ export class MarketplacePluginsService implements ICordovaPluginsService {
 		private $projectConstants: Project.IProjectConstants) { }
 
 	public getAvailablePlugins(): IFuture<any> {
-		return this.$server.cordova.getMarketplacePluginsData();
+		return this.$server.cordova.getMarketplacePluginsData(this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.Cordova);
```
rosen-vladimirov: maybe we could use 
```TypeScript
this.$project.projectData.Framework
```
This way we'll not change the code in the future, when we have nativescript plugins :)

lib/services/plugins-service.ts
```diff
@@ -49,6 +49,11 @@ export class PluginsService implements IPluginsService {
 				this.$errors.fail("No plugin name specified");
 			}
 
+			var pluginNameToLowerCase = pluginName.toLowerCase();
+			if(!_.any(this.getAvailablePlugins(), (pl) => pl.data.Name.toLowerCase() ===  pluginNameToLowerCase || pl.data.Identifier.toLowerCase() === pluginNameToLowerCase)) {
+				this.$errors.fail("Invalid plugin name: %s", pluginName);
```
rosen-vladimirov: `failWithoutHelp` seems more appropriate for me, it definitely looks like the user knows how to use the command, just the plugin id is incorrect.

rosen-vladimirov:
:+1:  two minor suggestions, but you can keep the current code as it is :)

PR #910, opened 2015-04-08
Add publish command

Mitko-Kerezov:
It enables publishing a **website** to a remote **ftp** server

Merge **after** https://github.com/telerik/mobile-cli-lib/pull/276 and https://github.com/Icenium/Ice/pull/3378 
Related to http://teampulse.telerik.com/view#item/286459

### DIFF
--- a/AppBuilderCLI.njsproj
+++ b/AppBuilderCLI.njsproj
@@ -49,6 +49,7 @@
     <TypeScriptCompile Include="lib\appbuilder-cli.ts" />
     <TypeScriptCompile Include="lib\bootstrap.ts" />
     <TypeScriptCompile Include="lib\commands\livepatch.ts" />
+    <TypeScriptCompile Include="lib\commands\publish.ts" />
     <TypeScriptCompile Include="lib\common\commands\preuninstall.ts" />
     <TypeScriptCompile Include="lib\config.ts" />
     <TypeScriptCompile Include="lib\declarations.d.ts" />
@@ -78,6 +79,7 @@
     <TypeScriptCompile Include="lib\services\generator-extensions-service.ts" />
     <TypeScriptCompile Include="lib\services\livepatch-service.ts" />
     <TypeScriptCompile Include="lib\services\project-commands-service.ts" />
+    <TypeScriptCompile Include="lib\services\publish-service.ts" />
     <TypeScriptCompile Include="lib\services\screen-builder-service.ts" />
     <TypeScriptCompile Include="lib\templates-service.ts" />
     <TypeScriptCompile Include="lib\x509.ts" />
--- a/docs/man_pages/publishing/publish-add.md
+++ b/docs/man_pages/publishing/publish-add.md
@@ -1,0 +1,26 @@
+publish add
+==========
+
+Usage | Synopsis
+------|-------
+General | `$ appbuilder publish add [<Name> [<PublishUrl>]]`
+
+Saves a new server connection. <% if(isHtml) { %>If you do not provide one or more command parameters, the Telerik AppBuilder CLI shows an interactive prompt to let you set the remaining details. <% } %>
+
+<% if(isConsole) { %>
+<% if(isCordova) { %>WARNING: This command is not applicable to Apache Cordova projects. To view the complete help for this command, run `$ appbuilder help publish add`<% } %>
+<% if(isNativeScript) { %>WARNING: This command is not applicable to NativeScript projects. To view the complete help for this command, run `$ appbuilder help publish add` <% } %>
+<% } %>
+
+<% if(isHtml) { %>
+### Command Limitations
+
+* You cannot run this command on NativeScript projects.
+* You cannot run this command on Apache Cordova projects.
+### Related Commands
+
+Command | Description
+----------|----------
+[publish](publish.html) | Lists all saved connections or publishes to the selected connection.
+[publish remove](publish-remove.html) | Removes a saved connection.
+<% } %>
--- a/docs/man_pages/publishing/publish-remove.md
+++ b/docs/man_pages/publishing/publish-remove.md
@@ -1,0 +1,30 @@
+publish remove
+==========
+
+Usage | Synopsis
+------|-------
+General | `$ appbuilder publish remove <Connection ID>`
+
+Removes a saved connection.
+
+<% if(isConsole) { %>
+<% if(isCordova) { %>WARNING: This command is not applicable to Apache Cordova projects. To view the complete help for this command, run `$ appbuilder help publish remove`<% } %>
+<% if(isNativeScript) { %>WARNING: This command is not applicable to NativeScript projects. To view the complete help for this command, run `$ appbuilder help publish remove` <% } %>
+<% } %>
+<% if((isConsole && isMobileWebsite) || isHtml) { %>
+### Attributes
+* `<Connection ID>` is the index or name of the connection as listed by `$ appbuilder publish`
+<% } %>
+<% if(isHtml) { %>
+### Command Limitations
+
+* You cannot run this command on NativeScript projects.
+* You cannot run this command on Apache Cordova projects.
+
+### Related Commands
+
+Command | Description
+----------|----------
+[publish](publish.html) | Lists all saved connections or publishes to the selected connection.
+[publish add](publish-add.html) | Saves a new server connection.
+<% } %>
--- a/docs/man_pages/publishing/publish.md
+++ b/docs/man_pages/publishing/publish.md
@@ -1,0 +1,42 @@
+publish
+==========
+
+Usage | Synopsis
+------|-------
+List connections | `$ appbuilder publish`
+Publish website to a saved connection | `$ appbuilder publish <Connection Index> [<Username> [<Password>]] [--force]`
+Publish website to URL | `$ appbuilder publish <URL> [<Username> [<Password>]] [--force]`
+Manage connections | `$ appbuilder publish <Command>`
+
+Lists saved server connections or lets you publish your website to a selected server connection. <% if(isHtml) { %>If you do not provide one or more command parameters, the Telerik AppBuilder CLI shows an interactive prompt to let you set the remaining details. <% } %>
+
+<% if(isConsole) { %>
+<% if(isCordova) { %>WARNING: This command is not applicable to Apache Cordova projects. To view the complete help for this command, run `$ appbuilder help publish`<% } %>
+<% if(isNativeScript) { %>WARNING: This command is not applicable to NativeScript projects. To view the complete help for this command, run `$ appbuilder help publish` <% } %>
+<% } %>
+
+<% if((isConsole && isMobileWebsite) || isHtml) { %>
+### Options
+* `--force | -f` - If set, purges the project directory on the remote server before uploading the mobile website.
+
+### Attributes
+* `<Connection Index>` is the index of the connection as listed by `$ appbuilder publish`
+* `<URL>` is the URL of a remote server.
+* `<Username>` and `<Password>` are your credentials for the remote server.
+* `<Command>` extends the `publish` command. You can set the following values for this attribute.
+	* `add` - Saves a new server connection.
+	* `remove` - Removes a saved server connection.
+<% } %>
+<% if(isHtml) { %>
+### Command Limitations
+
+* You cannot run this command on NativeScript projects.
+* You cannot run this command on Apache Cordova projects.
+
+### Related Commands
+
+Command | Description
+----------|----------
+[publish](publish.html) | Lists all saved connections or publishes to the selected connection.
+[publish add](publish-add.html) | Saves a new server connection.
+<% } %>
--- a/lib/commands/publish.ts
+++ b/lib/commands/publish.ts
@@ -1,0 +1,55 @@
+﻿///<reference path="../.d.ts"/>
+"use strict";
+
+import commandParams = require("../common/command-params");
+
+export class PublishListCommand implements ICommand {
+	constructor(private $injector: IInjector,
+		private $publishService: IPublishService) { }
+
+	allowedParameters: ICommandParameter[] = [new commandParams.StringCommandParameter(this.$injector), 
+		new commandParams.StringCommandParameter(this.$injector), new commandParams.StringCommandParameter(this.$injector)];
+
+	execute(args: string[]): IFuture<void> {
+		return (() => { 
+			if (args.length === 0) {
+					return this.$publishService.listAllConnections();
+			}
+
+			var idOrUrl = args[0];
+			var username = args[1];
+			var password = args[2];
+
+			this.$publishService.publish(idOrUrl, username, password).wait();
+		}).future<void>()();
+	}
+}
+$injector.registerCommand("publish|*list", PublishListCommand);
+
+export class PublishAddCommand implements ICommand {
+	constructor(private $injector: IInjector,
+		private $publishService: IPublishService) { }
+
+	allowedParameters: ICommandParameter[] = [new commandParams.StringCommandParameter(this.$injector), new commandParams.StringCommandParameter(this.$injector)];
+		
+	execute(args: string[]): IFuture<void> {
+		var name = args[0];
+		var publishUrl = args[1];
+		
+		return this.$publishService.addConnection(name, publishUrl);
+	}
+}
+$injector.registerCommand("publish|add", PublishAddCommand);
+
+export class PublishRemoveCommand implements ICommand {
+	constructor(private $publishService: IPublishService,
+		private $stringParameterBuilder: IStringParameterBuilder) { }
+
+	allowedParameters: ICommandParameter[] = [this.$stringParameterBuilder.createMandatoryParameter("Specify connection name or index.")];
+
+	execute(args: string[]): IFuture<void> {
+		var idOrName = args[0];
+		return this.$publishService.removeConnection(idOrName);
+	}
+}
+$injector.registerCommand("publish|remove", PublishRemoveCommand);
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -160,6 +160,7 @@ interface IProjectCapabilities {
 	livesyncCompanion: boolean;
 	updateKendo: boolean;
 	emulate: boolean;
+	publish: boolean;
 	uploadToAppstore: boolean;
 }
 
@@ -478,4 +479,18 @@ interface IDynamicSubCommandInfo {
 	baseCommandName: string;
 	filePath: string;
 	commandConstructor: Function;
+}
+
+interface IPublishService {
+	publish(idOrUrl: string, username: string, password: string): IFuture<void>;
+	listAllConnections(): void;
+	addConnection(name: string, publishUrl: string): IFuture<void>;
+	removeConnection(idOrName: string): IFuture<void>;
+}
+
+interface IPublishConnection extends IStringDictionary {
+	type: string;
+	publicUrl: string;
+	publishUrl: string;
+	name: string;
 }
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -24,12 +24,14 @@ var knownOpts: any = {
 		"release": Boolean,
 		"debug": Boolean,
 		"valid-value": Boolean,
-		"screenBuilderCacheDir": String
+		"screenBuilderCacheDir": String,
+		"force": Boolean
 	},
 	shorthands: IStringDictionary = {
 		"t": "template",
 		"r": "release",
-		"d": "debug"
+		"d": "debug",
+		"f": "force"
 	};
 
 _.extend(commonOptions.knownOpts, knownOpts);
--- a/lib/project/project-files-manager.ts
+++ b/lib/project/project-files-manager.ts
@@ -94,6 +94,14 @@ export class ProjectFilesManager implements Project.IProjectFilesManager {
 		return this.isFileExcluded(relativeToProjectPath, excludedProjectDirsAndFiles, projectDir);
 	}
 
+	public excludeFile(projectDir: string, excludeFilePath: string) : void {
+		if (!this.isProjectFileExcluded(projectDir, excludeFilePath)) {
+			var relativeToProjectPath = path.relative(projectDir, excludeFilePath);
+			var ignoreFilePath = path.join(projectDir, ProjectFilesManager.IGNORE_FILE);
+			this.$fs.appendFile(ignoreFilePath, relativeToProjectPath).wait();
+		}
+	}
+
 	private get ignoreFilesConfigurations(): string[] {
 		var configurations: string[] = [ ProjectFilesManager.IGNORE_FILE ];
 		// unless release is explicitly set, we use debug config
--- a/lib/project/project-files-manager.ts
+++ b/lib/project/project-files-manager.ts
@@ -97,8 +97,9 @@ export class ProjectFilesManager implements Project.IProjectFilesManager {
 	public excludeFile(projectDir: string, excludeFilePath: string) : void {
 		if (!this.isProjectFileExcluded(projectDir, excludeFilePath)) {
 			var relativeToProjectPath = path.relative(projectDir, excludeFilePath);
+			var appendData = '\n' + relativeToProjectPath + '\n';
 			var ignoreFilePath = path.join(projectDir, ProjectFilesManager.IGNORE_FILE);
-			this.$fs.appendFile(ignoreFilePath, relativeToProjectPath).wait();
+			this.$fs.appendFile(ignoreFilePath, appendData).wait();
 		}
 	}
 
--- a/lib/services/plugins-service.ts
+++ b/lib/services/plugins-service.ts
@@ -339,14 +339,7 @@ export class PluginsService implements IPluginsService {
 
 	private promptForVersionCore(pluginName: string, versions: any[]): IFuture<string> {
 		return (() => {
-			var promptData = [{
-				type: "list",
-				name: "version",
-				message: "Which plugin version would you like to use?",
-				choices: versions
-			}];
-			var answer = this.$prompter.get(promptData).wait();
-			var version = answer.version;
+			var version = this.$prompter.promptForChoice("Which plugin version do you want to use?", versions).wait();
 			return version;
 		}).future<string>()();
 	}
--- a/lib/services/publish-service.ts
+++ b/lib/services/publish-service.ts
@@ -1,0 +1,226 @@
+﻿///<reference path="../.d.ts"/>
+"use strict";
+
+import options = require("../common/options");
+import os = require("os");
+import helpers = require("../common/helpers");
+import path = require("path");
+import util = require("util");
+var Table = require("cli-table");
+
+export class PublishService implements IPublishService {
+	private static JSON_PUBLISH_FILE_NAME = ".abpublish";
+	private static ALLOWED_CONNECTION_TYPE = "ftp";
+	private static ERROR_MESSAGE_WHEN_PROJECT_EXISTS_ON_REMOTE = "User cancelled operation";
+	private static PUBLIC_URL_DEFAULT_VALUE = "";
+	private static REMOVE_ALL_CONNECTIONS_MESSAGE = "All of them";
+
+	private allPublishConnections: IPublishConnection[];
+	private ftpPublishConnections: IPublishConnection[];
+
+	constructor(private $errors: IErrors,
+				private $fs: IFileSystem,
+				private $logger: ILogger,
+				private $progressIndicator: IProgressIndicator,
+				private $projectConstants: Project.IProjectConstants,
+				private $projectFilesManager: Project.IProjectFilesManager,
+				private $project: Project.IProject,
+				private $prompter: IPrompter,
+				private $server: Server.IServer) {
+
+		this.$project.ensureProject();
+
+		if (!this.$project.capabilities.publish) {
+			this.$errors.failWithoutHelp("The publish command is only applicable to projects of type %s", this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.MobileWebsite);
+		}
+
+		this.readConnections().wait();
+		this.ftpPublishConnections = _.filter(this.allPublishConnections, c => c.type === PublishService.ALLOWED_CONNECTION_TYPE);
+	}
+
+	public publish(idOrUrl: string, username: string, password: string): IFuture<void> {
+		var ftpConnectionData = this.getFtpConnectionData(idOrUrl, username, password).wait();
+		return this.publishToRemote(ftpConnectionData);
+	}
+
+	public listAllConnections(): void {
+		if (!this.ftpPublishConnections.length) {
+			this.$logger.info("No connections found. To add a connection, run $ appbuilder publish add");
+			return;
+		}
+
+		var table = new Table({
+			head: ["Index", "Name", "Publish URL"],
+			chars: {'mid': '', 'left-mid': '', 'mid-mid': '', 'right-mid': ''}
+		});
+
+		_.forEach(this.ftpPublishConnections, (connection, index) => {
+			table.push([index + 1, connection.name, connection.publishUrl]);
+		});
+
+		this.$logger.out(table.toString());
+		this.$logger.info('To publish to a remote url you can run $ appbuilder publish <Connection Index> [<Username> [<Password>]] [--force]');
+	}
+
+	public addConnection(name: string, publishUrl: string): IFuture<void> {
+		return (() => {
+			var ftpPublishConnection: IPublishConnection = {
+				type: PublishService.ALLOWED_CONNECTION_TYPE,
+				publicUrl: PublishService.PUBLIC_URL_DEFAULT_VALUE,
+				publishUrl: publishUrl,
+				name: name
+			};
+
+			if (!ftpPublishConnection.name) {
+				ftpPublishConnection.name = this.$prompter.getString("Name:").wait();
+			}
+
+			if (!ftpPublishConnection.publishUrl) {
+				ftpPublishConnection.publishUrl = this.$prompter.getString("Publish URL:").wait();
+			}
+
+			if (_.findIndex(this.ftpPublishConnections, c => c.name === name && c.publishUrl === publishUrl) !== -1) {
+				this.$errors.failWithoutHelp("This connection has already been added.");
+			}
+
+			this.allPublishConnections.push(ftpPublishConnection);
+			this.savePublishJsonFile().wait();
+			this.$logger.info("Connection '%s' added successfully.", ftpPublishConnection.name);
+		}).future<void>()();
+	}
+
+	public removeConnection(idOrName: string): IFuture<void> {
+		return (() => {
+			var id = parseInt(idOrName) - 1;
+			var message : string;
+			if (!isNaN(id)) {
+				var removedConnection = this.removeConnectionById(id);
+				message = util.format("Connection '%s' successfully removed.", removedConnection.name);
+			} else {
+				var removedConnections = this.removeConnectionByName(idOrName).wait();
+				message = removedConnections.length > 1 ? util.format("All connections named '%s' successfully removed.", idOrName) : util.format("Connection '%s' successfully removed.", removedConnections[0].name);
+			}
+
+			this.savePublishJsonFile().wait();
+			this.$logger.info(message);
+		}).future<void>()();
+	}
+
+	private removeConnectionById(id: number): IPublishConnection {
+		this.validateId(id);
+
+		return _.first(helpers.remove(this.allPublishConnections, c => c === this.ftpPublishConnections[id]));
+	}
+
+	private removeConnectionByName(name: string): IFuture<IPublishConnection[]> {
+		return (():IPublishConnection[] => {
+			name = name.toLowerCase();
+			var connectionsToBeRemoved = _.filter(this.ftpPublishConnections, c => c.name.toLowerCase() === name);
+
+			if (connectionsToBeRemoved.length > 1) {
+				var connectionNameUrls = _.map(connectionsToBeRemoved, c => c.name + '(' + c.publishUrl + ')');
+				connectionNameUrls.push(PublishService.REMOVE_ALL_CONNECTIONS_MESSAGE);
+
+				var chosenConnectionNameUrl = this.$prompter.promptForChoice("Which connection do you want to remove?", connectionNameUrls).wait();
+				if (chosenConnectionNameUrl === PublishService.REMOVE_ALL_CONNECTIONS_MESSAGE) {
+					return <IPublishConnection[]>_.remove(this.allPublishConnections, c => c.name.toLowerCase() === name && c.type === PublishService.ALLOWED_CONNECTION_TYPE);
+				} else {
+					return helpers.remove(this.allPublishConnections, c => c.name + '(' + c.publishUrl + ')' === chosenConnectionNameUrl);
+				}
+			} else if (connectionsToBeRemoved.length === 1) {
+				return <IPublishConnection[]>_.remove(this.allPublishConnections, connectionsToBeRemoved[0]);
+			} else {
+				this.$errors.failWithoutHelp("Cannot find connections named '%s'.", name);
+			}
+		}).future<IPublishConnection[]>()();
+	}
+
+	private readConnections(): IFuture<void> {
+		return (() => {
+			var projectDir = this.$project.getProjectDir().wait();
+			var publishFilePath = path.join(projectDir, PublishService.JSON_PUBLISH_FILE_NAME);
+			if (!this.$fs.exists(publishFilePath).wait()) {
+				this.allPublishConnections = [];
+				return;
+			}
+
+			try {
+				this.allPublishConnections = this.$fs.readJson(publishFilePath).wait() || [];
+			} catch (err) {
+				this.$errors.failWithoutHelp("The publish file %s is corrupted." + os.EOL +
+						"Consider restoring an earlier version from your source control or backup." + os.EOL +
+						"Additional technical information: %s", publishFilePath, err.toString());
+			}
+		}).future<void>()();
+	}
+
+	private getFtpConnectionData(idOrUrl: string, username: string, password: string): IFuture<Server.FtpConnectionData> {
+		return (() => {
+			var ftpConnectionData: Server.FtpConnectionData = {
+				RemoteUrl: idOrUrl,
+				ShouldPurge: options.force,
+				Username: username,
+				Password: password
+			};
+
+			if (!ftpConnectionData.Username) {
+				ftpConnectionData.Username = this.$prompter.getString("Username:").wait();
+			}
+
+			if (!ftpConnectionData.Password) {
+				ftpConnectionData.Password = this.$prompter.getPassword("Password:", {allowEmpty: true}).wait();
+			}
+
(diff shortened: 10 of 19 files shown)


Fatme:
You should rebase the PR on master.

lib/declarations.d.ts
```diff
@@ -474,3 +475,17 @@ interface IProgressIndicator {
...
 }
+
+interface IPublishService {
+	publish(idOrUrl: string, username: string, password: string): IFuture<void>;
+	listAllConnections(): IFuture<void>;
+	addConnection(name: string, publishUrl: string): IFuture<void>;
+	removeConnection(idOrName: string): IFuture<void>;
+}
+
+interface IPublishConnection extends IDictionary<any> {
```
Fatme: `IStringDictionary` seems more suitable.

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+                private $project: Project.IProject,
+                private $prompter: IPrompter,
+                private $server: Server.IServer) {
+
+        if (!this.$project.capabilities.publish) {
+            this.$errors.failWithoutHelp("The publish command is only applicable to projects of type %s", this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.MobileWebsite);
+        }
+
+        this.readConnections().wait();
+        this.ftpPublishConnections = _.filter(this.allPublishConnections, c => c.type === PublishService.ALLOWED_COONECTION_TYPE);
```
Fatme: Shouldn't you add `this.$project.ensureProject()` ?

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+            this.$logger.printInfoMessageOnSameLine(os.EOL);
+            this.$logger.info("Project '%s' published successfully.", projectName);
+        }).future<void>()();
+    }
+
+    public listAllConnections(): IFuture<void> {
+        return (() => {
+            this.listConnections(this.ftpPublishConnections);
+
+            if (!this.ftpPublishConnections.length) {
```
Fatme: Shouldn't this check be inside of `listConnections` function?

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+            var message : string;
+            if (!isNaN(id)) {
+                this.validateId(id);
+
+                // (id - 1) is the index of the object to be deleted
+                _.remove(this.allPublishConnections, this.ftpPublishConnections[id - 1]);
+                message = util.format("Connection '%s' deleted successfully", this.ftpPublishConnections[id - 1].name);
+            } else {
+                var name = idOrName;
+                var connectionsToBeRemoved = _.filter(this.allPublishConnections, function(connection) {
```
Fatme: `_.filter(this.allPublishConnections, (connection) => connection.name === name)` is better

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+                    this.listConnections(connectionsToBeRemoved);
+                    var ftpConnectionId = parseInt(this.$prompter.getString("Index:").wait());
+                    if (isNaN(ftpConnectionId) || ftpConnectionId > connectionsToBeRemoved.length || ftpConnectionId < 0) {
+                        this.$errors.failWithoutHelp("Invalid index");
+                    }
+
+                    if (ftpConnectionId === 0) {
+                        _.remove(this.allPublishConnections, function(connection) {
+                                return connection.name === name;
+                            });
```
Fatme: ` _.remove(this.allPublishConnections, (connection) => connection.name === name);`

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+
+                    this.$errors.failWithoutHelp("The project file %s is corrupted." + os.EOL +
+                            "Consider restoring an earlier version from your source control or backup." + os.EOL +
+                            "Additional technical info: %s", publishFilePath, err.toString());
+                }
+            }
+        }).future<void>()();
+    }
+
+    private savePublishJsonFile(): void {
```
Fatme: This function should return future

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+
+            this.savePublishJsonFile();
+            this.$logger.info(message);
+        }).future<void>()();
+    }
+
+    private readConnections(): IFuture<void> {
+        return (() => {
+            var projectDir = this.$project.getProjectDir().wait();
+            if (projectDir) {
```
Fatme: If you add `this.$project.ensureProject()` in constructor before calling `readConnections` function, you should be able to remove this check :smile:

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+                var publishFilePath = path.join(projectDir, PublishService.JSON_PUBLISH_FILE_NAME);
+                try {
+                    this.allPublishConnections = this.$fs.readJson(publishFilePath).wait();
+                } catch (err) {
+                    if (err.code === 'ENOENT') {
```
Fatme: You can use `this.$fs.tryExecuteFileOperation`

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+                        return connection.name === name;
+                });
+
+                if (connectionsToBeRemoved.length > 1) {
+                    this.$logger.info("Multiple connections with name '%s' found." + os.EOL + 
+                                        "Type in the index of the one you want to delete " + os.EOL + 
+                                        "Or 0 if you want to delete them all", name);
+
+                    this.listConnections(connectionsToBeRemoved);
+                    var ftpConnectionId = parseInt(this.$prompter.getString("Index:").wait());
```
Fatme: Shouldn't we show all available connections instead of asking the user to enter the index of connection?

AppBuilderCLI.njsproj
```diff
@@ -47,6 +47,7 @@
     <Content Include="config\config.json" />
     <TypeScriptCompile Include="lib\appbuilder-cli.ts" />
     <TypeScriptCompile Include="lib\bootstrap.ts" />
+    <TypeScriptCompile Include="lib\commands\publish.ts" />
```
rosen-vladimirov: :smiley_cat:  someone else is using Visual Studio :smiley_cat:

docs/man_pages/publishing/publish-add.md
```diff
@@ -0,0 +1,18 @@
+publish add
+==========
+
+Usage | Syntax
+------|-------
+General | `$ appbuilder publish add [<Name>] [<PublishUrl>]`
```
rosen-vladimirov: is the Name mandatory parameter? What happens in case you do not pass it? This should be documented.

docs/man_pages/publishing/publish-remove.md
```diff
@@ -0,0 +1,19 @@
+publish
```
rosen-vladimirov: the title should be `publish-remove`

docs/man_pages/publishing/publish.md
```diff
@@ -0,0 +1,33 @@
+publish
+==========
+
+Usage | Syntax
+------|-------
+General | `$ appbuilder publish`    
+Additional functionality | `$ appbuilder publish [<Command>]`
+Publishing | `$ appbuilder publish [<Certificate ID>] [<Username>] [<Password>] [--force -f]`
+
+###General
```
rosen-vladimirov: shouldn't you have space after `### `-> `### General`

lib/options.ts
```diff
@@ -24,12 +24,14 @@ var knownOpts: any = {
...
-		"screenBuilderCacheDir": String
+		"screenBuilderCacheDir": String,
+		"force": Boolean
 	},
 	shorthands: IStringDictionary = {
 		"t": "template",
 		"r": "release",
-		"d": "debug"
+		"d": "debug",
+		"f": "force"
```
rosen-vladimirov: awesome :smile:

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+                ftpConnectionData.Username = this.$prompter.getString("Username:").wait();
+            }
+
+            if (!ftpConnectionData.Password) {
+                ftpConnectionData.Password = this.$prompter.getPassword("Password:", {allowEmpty: true}).wait();
+            }
+
+            var id = parseInt(idOrUrl);
+
+            // the check wether it contains a dot is performed
```
rosen-vladimirov: the chech __whether__

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+            // in case the user tries to publish to an IP address
+            if (!isNaN(id) && idOrUrl.indexOf('.') < 0) {
+                this.validateId(id);
+
+                // (id - 1) is the index of the object to be deleted
```
rosen-vladimirov: the comment says that the connection will be deleted, but the method is called publish. What are we going to do with this connection? :smirk_cat:

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+            this.$project.importProject().wait();
+            this.$logger.printInfoMessageOnSameLine("Publishing");
+
+            try {
+                this.$progressIndicator.showProgressIndicator(this.$server.publish.publishToFtp(projectName, projectName, ftpConnectionData), 6000).wait();
+            } catch (err) {
+                this.$logger.printInfoMessageOnSameLine(os.EOL);
+                if (err.message === PublishService.ERROR_MESSAGE_WHEN_PROJECT_EXISTS_ON_REMOTE) {
+                    this.$errors.failWithoutHelp("This project already exists on the remote server. " + os.EOL + 
+                                    "To remove all files and upload the new ones on the remote run --force option");
```
rosen-vladimirov: Maybe it is better to tell the exact command, or at least - `run the command with --force option`

lib/services/publish-service.ts
Mitko-Kerezov: With `this.$fs.tryExecuteFileOperation` I won't have access to `err.toString()` which provides invaluable information in case the json project file is messed up

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+"use strict";
+
+import options = require("../common/options");
+import os = require("os");
+import path = require("path");
+import util = require("util");
+
+export class PublishService implements IPublishService {
+    private static JSON_PUBLISH_FILE_NAME = ".abpublish";
+    private static ALLOWED_COONECTION_TYPE = "ftp";
```
rosen-vladimirov: `ALLOWED_CONNECTION_TYPE` maybe

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+            } else {
+                var name = idOrName;
+                var connectionsToBeRemoved = _.filter(this.allPublishConnections, function(connection) {
+                        return connection.name === name;
+                });
+
+                if (connectionsToBeRemoved.length > 1) {
+                    this.$logger.info("Multiple connections with name '%s' found." + os.EOL + 
+                                        "Type in the index of the one you want to delete " + os.EOL + 
+                                        "Or 0 if you want to delete them all", name);
```
rosen-vladimirov: or

lib/services/publish-service.ts
rosen-vladimirov: Nice idea, but will it be possible to add option - remove them all?

lib/services/publish-service.ts
rosen-vladimirov: In fact you can omit the '(' and ')':
```TypeScript
_.remove(this.allPublishConnections, connection => connection.name === name);
```

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+                    var ftpConnectionId = parseInt(this.$prompter.getString("Index:").wait());
+                    if (isNaN(ftpConnectionId) || ftpConnectionId > connectionsToBeRemoved.length || ftpConnectionId < 0) {
+                        this.$errors.failWithoutHelp("Invalid index");
+                    }
+
+                    if (ftpConnectionId === 0) {
+                        _.remove(this.allPublishConnections, function(connection) {
+                                return connection.name === name;
+                            });
+                        message = util.format("Connections named '%s' deleted successfully", name);
```
rosen-vladimirov: `.` at the end

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+
+            this.allPublishConnections.push(ftpPublishConnection);
+            this.savePublishJsonFile();
+            this.$logger.info("Connection '%s' added successfully.", ftpPublishConnection.name);
+        }).future<void>()();
+    }
+
+    public removeConnection(idOrName: string): IFuture<void> {
+        return (() => {
+            var id = parseInt(idOrName);
```
rosen-vladimirov:
```TypeScript
var id = parseInt(idOrName) - 1;
```
This way you can use id directly, without the need of ```id - 1``` everywhere

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    private savePublishJsonFile(): void {
+        var projectDir = this.$project.getProjectDir().wait();
+        this.$fs.writeJson(path.join(projectDir, PublishService.JSON_PUBLISH_FILE_NAME), this.allPublishConnections).wait();
+    }
+
+    private listConnections(connections: IPublishConnection[]): void {
+        _.forEach(connections, (connection, index) => {
+            this.$logger.out("%s:\tName: %s " + os.EOL + "\tPublishUrl: %s " + os.EOL,
```
rosen-vladimirov:
```TypeScript
 this.$logger.out("%s:\tName: %s %s \tPublishUrl: %s %s", 
                (index + 1).toString(), 
                os.EOL,
                connection.name,
                connection.publishUrl,
                os.EOL);
```

lib/services/publish-service.ts
rosen-vladimirov: also it should be: 
```TypeScript
this.allPublishConnections = _.remove(this.allPublishConnections, connection => connection.name === name);
```

lib/services/publish-service.ts
```diff
@@ -0,0 +1,211 @@
...
+                    }
+
+                    if (ftpConnectionId === 0) {
+                        _.remove(this.allPublishConnections, function(connection) {
+                                return connection.name === name;
+                            });
+                        message = util.format("Connections named '%s' deleted successfully", name);
+                    } else {
+                        // again (id - 1) is the index of the object to be deleted
+                        _.remove(this.allPublishConnections, connectionsToBeRemoved[ftpConnectionId - 1]);
```
rosen-vladimirov: I think remove operation is not correct here. First of all it doesn't modify the original array, so you should add `this.allPublishConnection = ` at the beginning. Also Remove operation with object is a little bit tricky, for example consider the following script:
```JavaScript
var b = ["a", "b", "c", "a"];
var a = _.remove(b, b[0])
console.log("b is: " + b);
console.log("a is " + a);
```

The output is:
```
b is: a,b,c,a
a is
```

lib/services/publish-service.ts
rosen-vladimirov: `getProjectDir` is also "waited", so it should be
```TypeScript
private savePublishJsonFile(): IFuture<void> {
    return (() => { 
         var projectDir = this.$project.getProjectDir().wait();
         this.$fs.writeJson(path.join(projectDir, PublishService.JSON_PUBLISH_FILE_NAME), this.allPublishConnections).wait();
    }).future<void>()()
}

File: lib/services/publish-service.ts (hunk 22 above)
Comment (rosen-vladimirov) [inline lib/services/publish-service.ts]: Also you can use cli-table for stunning UI in the console

File: lib/services/publish-service.ts [hunk 24]
```diff
@@ -0,0 +1,211 @@
...
+
+            if (!ftpPublishConnection.name) {
+                ftpPublishConnection.name = this.$prompter.getString("Name:").wait();
+            }
+
+            if (!ftpPublishConnection.publishUrl) {
+                ftpPublishConnection.publishUrl = this.$prompter.getString("Publish URL:").wait();
+            }
+
+            this.allPublishConnections.push(ftpPublishConnection);
```
rosen-vladimirov: What happens in case we already have the same connection in our file? At least a warning is required (IMO we should not allow such adding).

rosen-vladimirov: Ping @teobugslayer , @Fatme , @ligaz  - what do you think - is it correct to have same connections in the .abpublish file, I mean totally same objects with same name, url, etc.

docs/man_pages/publishing/publish.md
```diff
@@ -0,0 +1,35 @@
+publish
+==========
+
+Usage | Syntax
+------|-------
+General | `$ appbuilder publish`    
+Additional functionality | `$ appbuilder publish [<Command>]`
+Publishing | `$ appbuilder publish [<Connection ID>] [<Username>] [<Password>] [--force -f]`
```
rosen-vladimirov: only --force should be shown here

lib/services/plugins-service.ts
```diff
@@ -340,14 +340,8 @@ export class PluginsService implements IPluginsService {
...
-			var promptData = [{
-				type: "list",
-				name: "version",
-				message: "Which plugin version would you like to use?",
-				choices: versions
-			}];
-			var answer = this.$prompter.get(promptData).wait();
-			var version = answer.version;
+			var version = this.$prompter.promptForChoice("Which plugin version would you like to use?", versions).wait();
+			console.log(version);
```
rosen-vladimirov: :crying_cat_face:  maybe used for debugging only?

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+
+                throw err;
+            }
+
+            this.$logger.printInfoMessageOnSameLine(os.EOL);
+            this.$logger.info("Project '%s' published successfully.", projectName);
+        }).future<void>()();
+    }
+
+    public listAllConnections(): IFuture<void> {
```
rosen-vladimirov: why do we need this method and why the return type is future?

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+            var message : string;
+            if (!isNaN(id)) {
+                this.validateId(id);
+
+                var connectionIndex = _.findIndex(this.allPublishConnections, this.ftpPublishConnections[id]);
+                this.allPublishConnections.splice(connectionIndex, 1);
+                message = util.format("Connection '%s' successfully removed.", this.ftpPublishConnections[id].name);
+            } else {
+                var name = idOrName;
+                var connectionsToBeRemoved = _.filter(this.ftpPublishConnections, c => c.name === name);
```
rosen-vladimirov: do we want case-sensitive or case-insensitive match?

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+
+                if (connectionsToBeRemoved.length > 1) {
+                    var connectionNameUrls = _.map(connectionsToBeRemoved, c => c.name + '(' + c.publishUrl + ')');
+                    connectionNameUrls.splice(connectionNameUrls.length, 0, PublishService.REMOVE_ALL_CONNECTIONS_MESSAGE);
```
rosen-vladimirov: maybe `push` or `unshift` (based on which position you think it is better to add the option) as it is more readable for me (this is just a suggestion)

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+
+        if (!this.$project.capabilities.publish) {
+            this.$errors.failWithoutHelp("The publish command is only applicable to projects of type %s", this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.MobileWebsite);
+        }
+
+        this.readConnections().wait();
+        this.ftpPublishConnections = _.filter(this.allPublishConnections, c => c.type === PublishService.ALLOWED_CONNECTION_TYPE);
+    }
+
+    public publish(idOrUrl: string, username: string, password: string): IFuture<void> {
```
teobugslayer: You may use string|int type and give id as int here

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+            var publishFilePath = path.join(projectDir, PublishService.JSON_PUBLISH_FILE_NAME);
+            try {
+                this.allPublishConnections = this.$fs.readJson(publishFilePath).wait();
+            } catch (err) {
+                if (err.code === 'ENOENT') {
+                    this.allPublishConnections = [];
+                    return;
+                }
+
+                this.$errors.failWithoutHelp("The project file %s is corrupted." + os.EOL +
```
teobugslayer: This error message mentions the project file, but you are dealing with the publish file. I think it should be reworded.

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+                private $prompter: IPrompter,
+                private $server: Server.IServer) {
+
+        this.$project.ensureProject();
+
+        if (!this.$project.capabilities.publish) {
+            this.$errors.failWithoutHelp("The publish command is only applicable to projects of type %s", this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.MobileWebsite);
+        }
+
+        this.readConnections().wait();
```
teobugslayer: this can throw ($errors.fail actually throws). Because this is a constructor, yok would barf its unhappiness upon the user in the form of "Cannot construct blah-blah" message. Is this OK with you?

lib/services/publish-service.ts
teobugslayer: This function has an UI and makes server calls at the same time. Consider splitting it into smaller, more focused parts.

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+
+            var projectDir = this.$project.getProjectDir().wait();
+            this.$projectFilesManager.excludeFile(projectDir, PublishService.JSON_PUBLISH_FILE_NAME);
+            this.$project.importProject().wait();
+            this.$logger.printInfoMessageOnSameLine("Publishing");
+
+            try {
+                this.$progressIndicator.showProgressIndicator(this.$server.publish.publishToFtp(projectName, projectName, ftpConnectionData), 6000).wait();
+            } catch (err) {
+                this.$logger.printInfoMessageOnSameLine(os.EOL);
```
teobugslayer: Isn't $logger.out() the same?

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+                if (err.message === PublishService.ERROR_MESSAGE_WHEN_PROJECT_EXISTS_ON_REMOTE) {
+                    this.$errors.failWithoutHelp("This project already exists on the remote server. " + os.EOL + 
+                                    "To remove all files and upload the new ones on the remote run $ appbuilder publish remove --force");
```
teobugslayer: "on the remote" - on remote what?

lib/services/publish-service.ts
teobugslayer: I agree, we need to check for this and warn the user.

rosen-vladimirov: Warn or prevent him from writing such data in the file?

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+                ftpPublishConnection.publishUrl = this.$prompter.getString("Publish URL:").wait();
+            }
+
+            this.allPublishConnections.push(ftpPublishConnection);
+            this.savePublishJsonFile().wait();
+            this.$logger.info("Connection '%s' added successfully.", ftpPublishConnection.name);
+        }).future<void>()();
+    }
+
+    public removeConnection(idOrName: string): IFuture<void> {
```
teobugslayer: This method is too long and complicated - consider splitting it

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+    private savePublishJsonFile(): IFuture<void> {
+        return (() => { 
+            var projectDir = this.$project.getProjectDir().wait();
+            this.$fs.writeJson(path.join(projectDir, PublishService.JSON_PUBLISH_FILE_NAME), this.allPublishConnections).wait();
+        }).future<void>()()
+    }
+
+    private listConnections(connections: IPublishConnection[]): void {
+        if (!connections.length) {
+            this.$logger.info("No connections found. To add a connection, run `connection add`");
```
teobugslayer: The help message needs rewording. ``` run '$appbuilder connection add' ```

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+
+        var table = new Table({
+                head: ["Index", "Name", "Publish URL"]
+            });
+
+        _.forEach(connections, (connection, index) => {
+            table.push([index + 1, connection.name, connection.publishUrl]);
+        });
+
+        this.$logger.out(table.toString());
```
teobugslayer: Consider removing the middle lines in the table to improve information density and readability. See https://www.npmjs.com/package/cli-table#custom-styles

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    private validateId(id: number): void {
+        if (this.ftpPublishConnections.length <= id) {
+            this.$errors.failWithoutHelp("You only have %s connections available", this.ftpPublishConnections.length);
```
teobugslayer: I think this message is not descriptive enough. What should the user do to fix the cause of this message? @ikoevska , your opinion?

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+"use strict";
+
+import options = require("../common/options");
+import os = require("os");
+import path = require("path");
+import util = require("util");
+var Table = require("cli-table");
+
+export class PublishService implements IPublishService {
+    private static JSON_PUBLISH_FILE_NAME = ".abpublish";
```
teobugslayer: It would be great to add some basic unit tests for this service

lib/services/publish-service.ts
teobugslayer: Indeed, listConnections without parameters can do the same

lib/services/publish-service.ts
Fatme: Actually it was strange and confusing to me but after talking with @mitko in person we decided that we need to support this case.  Actually we may have two connnections with the same name and url but with different site url. So I think it is better to show a warning.

rosen-vladimirov: @Fatme , for the case you are talking about, it is ok to have such connections. But currently we allow adding exactly the same record in .abpublish file, which is my pain...

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    private readConnections(): IFuture<void> {
+        return (() => {
+            var projectDir = this.$project.getProjectDir().wait();
+            var publishFilePath = path.join(projectDir, PublishService.JSON_PUBLISH_FILE_NAME);
+            try {
+                this.allPublishConnections = this.$fs.readJson(publishFilePath).wait();
+            } catch (err) {
+                if (err.code === 'ENOENT') {
```
ligaz: It is better to check if the file exist rather than catching errors.

lib/services/publish-service.ts
ligaz: Isn't it publish add?

docs/man_pages/publishing/publish.md
```diff
@@ -0,0 +1,35 @@
...
+
+### General
+  Lists all saved connections
```
ikoevska: Doesn't this command actually upload rather than list?

docs/man_pages/publishing/publish-add.md
ikoevska: I am slightly concerned that this publish command might cause confusion in users who have been using our other clients. They might be confused that we have a publish set of commands, appstore set of commands and appmanager set of commands. I am wondering whether the publish set of commands should be redesigned to nest the other commands?

In addition, it might be a good idea to rename the server connection set of commands to connection or something. And have a publish website command for the upload?

lib/services/plugins-service.ts
ikoevska: How is this related to the publish command? :blush:

Anyway "Which plugin version do you want to use?"

lib/services/publish-service.ts
ikoevska: To remove all files and upload the new ones on the remote server, run $ appbuilder publish remove --force

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+                if (err.message === PublishService.ERROR_MESSAGE_WHEN_PROJECT_EXISTS_ON_REMOTE) {
+                    this.$errors.failWithoutHelp("This project already exists on the remote server. " + os.EOL + 
+                                    "To remove all files and upload the new ones on the remote run $ appbuilder publish remove --force");
+                }
+
+                throw err;
+            }
+
+            this.$logger.printInfoMessageOnSameLine(os.EOL);
+            this.$logger.info("Project '%s' published successfully.", projectName);
```
ikoevska: Is it published or uploaded? I am almost certain that uploaded is the more correct term here.

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+                message = util.format("Connection '%s' successfully removed.", this.ftpPublishConnections[id].name);
+            } else {
+                var name = idOrName;
+                var connectionsToBeRemoved = _.filter(this.ftpPublishConnections, c => c.name === name);
+
+                if (connectionsToBeRemoved.length > 1) {
+                    var connectionNameUrls = _.map(connectionsToBeRemoved, c => c.name + '(' + c.publishUrl + ')');
+                    connectionNameUrls.splice(connectionNameUrls.length, 0, PublishService.REMOVE_ALL_CONNECTIONS_MESSAGE);
+
+                    var chosenConnectionNameUrl = this.$prompter.promptForChoice("Which connection would you like to remove?", connectionNameUrls).wait();
```
ikoevska: Which connection do you want to remove?

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+                    } else {
+                        var chosenConnectionIndex = _.findIndex(this.allPublishConnections, c => c.name + '(' + c.publishUrl + ')' === chosenConnectionNameUrl);
+                        this.allPublishConnections.splice(chosenConnectionIndex, 1);
+                        message = util.format("Connection '%s' successfully removed.", chosenConnectionNameUrl);
+                    }
+                } else if (connectionsToBeRemoved.length === 1) {
+                    _.remove(this.allPublishConnections, connectionsToBeRemoved[0]);
+                    message = util.format("Connection '%s' successfully removed.", connectionsToBeRemoved[0].name);
+                } else {
+                    this.$errors.failWithoutHelp("No connections named '%s' found.", name);
```
ikoevska: This wording is fine but we usually go with "Cannot find..." In this case it would be "Cannot find connections named X."

lib/services/publish-service.ts
```diff
@@ -0,0 +1,208 @@
...
+                this.allPublishConnections = this.$fs.readJson(publishFilePath).wait();
+            } catch (err) {
+                if (err.code === 'ENOENT') {
+                    this.allPublishConnections = [];
+                    return;
+                }
+
+                this.$errors.failWithoutHelp("The project file %s is corrupted." + os.EOL +
+                        "Consider restoring an earlier version from your source control or backup." + os.EOL +
+                        "Additional technical info: %s", publishFilePath, err.toString());
```
ikoevska: information

lib/services/publish-service.ts
ikoevska: I am guessing this is a validation message when the user sets the connection by ID? And the ID exceeds the number of stored connections? Just some random ramblings - but what happens when the user specifies an invalid ID, say 0 or a negative number? Or not an integer?

If I've figured this out correctly, you can show the following message. "Incorrect ID. For a complete list of the available connections and their IDs, run $ appbuilder publish list"

lib/services/plugins-service.ts
Mitko-Kerezov: This is a side effect of some refactoring in the prompter.

lib/services/publish-service.ts
teobugslayer: I disagree, a file system entity may exist but is unusable. Typical examples are - a directory with the same name, file is being locked (usually a bad anti-virus, but text editors are also known to do this), there may be attributes or ACLs which deny reading the file. One must write the try / catch anyway so the exists check is unneeded.

docs/man_pages/publishing/publish.md
Mitko-Kerezov: If called without parameters it acts as `$ appbuilder publish list`, which is reflected in documentation.
After a discussion with @rosen-vladimirov it was decided that `$ appbuilder publish list` should not be included in the docs

lib/services/publish-service.ts
Mitko-Kerezov: I agree about the term, however we *upload* the project to our server and after that *publish* to the remote server - this is why I chose this particular wording

lib/services/publish-service.ts
Mitko-Kerezov: `$logger.out` can't be used without parameters so I guess it's more sensible the way it is now rather than `$logger.out('')` or something like that

Mitko-Kerezov:
Comments addressed, ping @Fatme, @rosen-vladimirov and @teobugslayer

lib/services/publish-service.ts
teobugslayer: This is a bug in typescript definitions :)

Try this code

```
			this.$logger.out("A");
			this.$logger.out();
			this.$logger.out("A");
```

tsc will complain that there's an error, but when you run the program, the expected empty line will be printed out.

lib/services/publish-service.ts
```diff
@@ -0,0 +1,226 @@
...
+            this.$logger.info("No connections found. To add a connection, run $ appbuilder publish add");
+            return;
+        }
+
+        var table = new Table({
+            head: ["Index", "Name", "Publish URL"],
+            chars: {'mid': '', 'left-mid': '', 'mid-mid': '', 'right-mid': ''}
+        });
+
+        _.forEach(this.ftpPublishConnections, (connection, index) => {
```
rosen-vladimirov: _.forEach + push inside the each... this sounds like _.map

lib/services/publish-service.ts
```diff
@@ -0,0 +1,226 @@
...
+
+            if (!ftpPublishConnection.name) {
+                ftpPublishConnection.name = this.$prompter.getString("Name:").wait();
+            }
+
+            if (!ftpPublishConnection.publishUrl) {
+                ftpPublishConnection.publishUrl = this.$prompter.getString("Publish URL:").wait();
+            }
+
+            if (_.findIndex(this.ftpPublishConnections, c => c.name === name && c.publishUrl === publishUrl) !== -1) {
```
rosen-vladimirov: why `_.findIndex` and not just `_.find`:
```TypeScript
if (_.find(this.ftpPublishConnections, c => c.name === name && c.publishUrl === publishUrl))
```

lib/services/publish-service.ts
Mitko-Kerezov: I don't think _.map is applicable here because within the ._forEach in every iteration I push a small *array* whereas _.map would return an array containing all of the small arrays and I'd still have to iterate it to get its values

lib/commands/publish.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export class PublishListCommand implements ICommand {
+    constructor(private $injector: IInjector,
+        private $publishService: IPublishService) { }
+
+    allowedParameters: ICommandParameter[] = [new commandParams.StringCommandParameter(this.$injector), 
+        new commandParams.StringCommandParameter(this.$injector), new commandParams.StringCommandParameter(this.$injector)];
+
+    execute(args: string[]): IFuture<void> {
+        if (args.length === 0) {
```
teobugslayer: This if looks very peculiar. Can we just remove it and make publish|list default command?

lib/project/project-files-manager.ts
```diff
@@ -94,6 +94,14 @@ export class ProjectFilesManager implements Project.IProjectFilesManager {
 		return this.isFileExcluded(relativeToProjectPath, excludedProjectDirsAndFiles, projectDir);
 	}
 
+	public excludeFile(projectDir: string, excludeFilePath: string) : void {
+		if (!this.isProjectFileExcluded(projectDir, excludeFilePath)) {
+			var relativeToProjectPath = path.relative(projectDir, excludeFilePath);
+			var ignoreFilePath = path.join(projectDir, ProjectFilesManager.IGNORE_FILE);
+			this.$fs.appendFile(ignoreFilePath, relativeToProjectPath).wait();
```
teobugslayer: This would break the .abignore file, if the last existing line in it does not end with a new line char

lib/services/publish-service.ts
```diff
@@ -0,0 +1,226 @@
...
+            this.savePublishJsonFile().wait();
+            this.$logger.info(message);
+        }).future<void>()();
+    }
+
+    private removeConnectionById(id: number): IPublishConnection {
+        this.validateId(id);
+
+        var connectionIndex = _.findIndex(this.allPublishConnections, this.ftpPublishConnections[id]);
+        return _.first(this.allPublishConnections.splice(connectionIndex, 1));
```
teobugslayer: Untested, but what about
```
return _.(this.allPublishConnections)
	.pullAt(connectionIndex)
	.first()
	.value();
```?

File: lib/services/publish-service.ts [hunk 51]
```diff
@@ -0,0 +1,226 @@
...
+            this.$projectFilesManager.excludeFile(projectDir, PublishService.JSON_PUBLISH_FILE_NAME);
+            this.$project.importProject().wait();
+            this.$logger.printInfoMessageOnSameLine("Publishing");
+
+            try {
+                this.$progressIndicator.showProgressIndicator(this.$server.publish.publishToFtp(projectName, projectName, ftpConnectionData), 6000).wait();
+            } catch (err) {
+                this.$logger.out();
+                if (err.message === PublishService.ERROR_MESSAGE_WHEN_PROJECT_EXISTS_ON_REMOTE) {
+                    this.$errors.failWithoutHelp("This project already exists on the remote server. " + os.EOL +
```
teobugslayer: failWithoutHelp(os.EOL + "This project ......

teobugslayer:
:+1: after rebasing and green build.

rosen-vladimirov:
:+1:  from me as well, great job :smile_cat:

ligaz:
run ci

lib/services/publish-service.ts
Mitko-Kerezov: `this.$logger.out()` is needed because regardless of what error is thrown we don't want to print the message on the same line

lib/services/publish-service.ts
Fatme: `_.map would return an array containing all of the small arrays and I'd still have to iterate it to get its values` - You can _.flatten to get the values.

Mitko-Kerezov: `_.map` would return something like this: `[[1, 'string1'], [2, 'string2'], ..]`
What I need is: `[1, 'string1'], [2, 'string2'], ..` - without the outer array
What I get with `_.flatten` is `[1, 'string1', 2, 'string2']`

rosen-vladimirov:
:+1:  for the minor bug fix commit

Fatme:
:+1:

docs/man_pages/publishing/publish-add.md
```diff
@@ -0,0 +1,21 @@
+publish add
+==========
+
+Usage | Syntax
```
rosen-vladimirov: Replace Syntax with Synopsis in each of your newly added articles.

docs/man_pages/publishing/publish-add.md
```diff
@@ -0,0 +1,21 @@
+publish add
```
ikoevska: publish add
==========

Usage | Synopsis
------|-------
General | `$ appbuilder publish add [<Name> [<PublishUrl>]]`

Saves a new server connection. \<% if(isHtml) { %\>If you do not provide one or more command parameters, the Telerik AppBuilder CLI shows an interactive prompt to let you set the remaining details. 

\#\#\# Related Commands

Command | Description
----------|----------
[publish](publish.html) | Lists all saved connections or publishes to a connection/remote ftp server
[publish remove](publish-remove.html) | Removes a saved connection
\<% } %\>

docs/man_pages/publishing/publish-remove.md
```diff
@@ -0,0 +1,19 @@
+publish remove
```
ikoevska: publish remove
==========

Usage \| Synopsis
------\|-------
General \| \`$ appbuilder publish remove [\<Connection ID\>]\`

Removes a saved connection. If you do not provide one or more command parameters, the Telerik AppBuilder CLI shows an interactive prompt to let you set the remaining details. 

\#\#\# Attributes
* \`\<Connection ID\>\` is the index or name of the connection as listed by \`$ appbuilder publish`

\<% if(isHtml) { %\>
\#\#\# Related Commands

Command \| Description
----------\|----------
\[publish\]\(publish.html\) \| Lists all saved connections or publishes to the selected connection.
\[publish add\]\(publish-add.html\) \| Saves a new server connection.
\<% } %\>

docs/man_pages/publishing/publish.md
```diff
@@ -0,0 +1,36 @@
+publish
+==========
+
+Usage | Synopsis
```
ikoevska: Usage \| Synopsis
------\|-------
List connections \| \`$ appbuilder publish\`
Publish website to a saved connection \| \`$ appbuilder publish \<Connection ID\> [\<Username\> [\<Password\>]] [--force]`
Publish website to URL \| \`$ appbuilder publish \<URL\> [\<Username\> [\<Password\>]] [--force]`
Manage connections \| \`$ appbuilder publish \<Command\>\`

Lists saved server connections or lets you publish your website to a selected server connection. \<% if(isHtml) { %\>If you do not provide one or more command parameters, the Telerik AppBuilder CLI shows an interactive prompt to let you set the remaining details. \<% } %\> 

\<% if(isConsole) { %\>
\<% if(isCordova) { %\>WARNING: This command is not applicable to Apache Cordova projects. To view the complete help for this command, run `$ appbuilder help publish`\<% } %\>
\<% if(isNativeScript) { %\>WARNING: This command is not applicable to NativeScript projects. To view the complete help for this command, run `$ appbuilder help publish` \<% } %\>
\<% } %\> 

<% if((isConsole && isMobileWebsite) || isHtml) { %>  
\#\#\# Options
\* \`--force | -f\` - If set, purges the project directory on the remote server before uploading the mobile website.


\#\#\# Attributes
\* \`\<Connection ID\>\` is the index of the connection as listed by \`$ appbuilder publish`
\* \`\<URL\>\` is the URL of a remote server.
\* \`\<Username\>\` and \`\<Password\>\` are your credentials for the remote server.
\* \`\<Command\>` extends the `publish` command. You can set the following values for this attribute.
	\* \`add\` - Saves a new server connection.
	\* \`remove\` - Removes a saved server connection.

\<% } %\>
\<% if(isHtml) { %\>
\#\#\# Command Limitations

* You cannot run this command on NativeScript projects.
* You cannot run this command on Apache Cordova projects.

\#\#\# Related Commands

Command \| Description
----------\|----------
\[publish\]\(publish.html\) \| Lists all saved connections or publishes to the selected connection.
\[publish add\]\(publish-add.html\) \| Saves a new server connection.
\<% } %\>

PR #912, opened 2015-04-08
Add timeout on preuninstall

rosen-vladimirov:
In some cases the preuninstall operations requires some more time and leads to errors if we do exit immediately. Add timeout of 3 seconds (it was removed with one of the previous commits by mistake).
Sync with latest common lib.


Fatme:
:+1:

PR #913, opened 2015-04-08
Add command that lists all supported SB commands

Fatme:


lib/commands/screenbuilder.ts
```diff
@@ -0,0 +1,18 @@
...
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			var allSupportedCommands = this.$screenBuilderService.allSupportedCommands().wait();
+			_.each(allSupportedCommands, (command, index) => this.$logger.out("%s: %s", index.toString(), command));
+		}).future<void>()();
+	}
+
+	allowedParameters: ICommandParameter[] = [];
+}
+$injector.registerCommand("screenbuilder", ScreenBuilderCommand);
```
Fatme: I couldn't figure out better name for this command, so I'm open to any suggestion.

lib/commands/screenbuilder.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class ScreenBuilderCommand implements ICommand {
+
+	constructor(private $logger: ILogger,
+		private $screenBuilderService: IScreenBuilderService) { }
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			var allSupportedCommands = this.$screenBuilderService.allSupportedCommands().wait();
+			_.each(allSupportedCommands, (command, index) => this.$logger.out("%s: %s", index.toString(), command));
```
rosen-vladimirov: why do we need the index here? What's the value of showing it to the users?

rosen-vladimirov:
run ci

lib/commands/screenbuilder.ts
Fatme: I think it is better to be enumerated :smile_cat:

teobugslayer:
We talked in person and decided to put the SB commands in the text of the help command.

rosen-vladimirov:
:+1:

PR #914, opened 2015-04-09
Fix 'grunt test' task

Mitko-Kerezov:
It now compiles and runs all tests


rosen-vladimirov:
:+1:  after green build :)

PR #917, opened 2015-04-14
Add IPrompter methods to unit test stubs

Mitko-Kerezov:
So that our mocks correctly implement the interface


rosen-vladimirov:
:+1:  after green (or almost green) build

PR #918, opened 2015-04-14
AppManager LiveSync

rosen-vladimirov:
Allow publishing of patches in AppManager. Add new command: `appmanager livesync`. It can be used without parameters in which case the user must confirm that the patch should be published to all platforms. The command accepts multiple arguments, which are the platforms. It can be used with one or with many platforms:
* `$ appmanager livesync android`
* `$ appmanager livesync android ios`
etc.

In case LivePatch plugin is not enabled for release configuration, we notify the user the we are enabling it.
http://teampulse.telerik.com/view#/12319

### DIFF
--- a/docs/man_pages/publishing/appmanager-livesync.md
+++ b/docs/man_pages/publishing/appmanager-livesync.md
@@ -1,21 +1,36 @@
-certificate
+appmanager livesync
 ==========
 
 Usage | Synopsis
 ------|-------
-List certificates | `$ appbuilder certificate`    
-Manage certificates | `$ appbuilder certificate [<Command>]`
+General | `$ appbuilder appmanager livesync [<Platforms>]`
 
-Lists or lets you manage certificates for code signing iOS and Android applications. <% if(isHtml) { %>When building an app, you can set the certificate by index or name in the `--certificate` option.<% } %>  
+Publish a new update of your application in AppManager. <% if(isHtml) { %>If you have not enabled the AppManager LiveSync plugin for your project, it will be automatically enabled for the release build configuration.<% } %>
+In case you do not specify a platform, the patch will be published for all platforms (in case you confirm the prompt dialog).
 
+<% if(isConsole && isMobileWebsite) { %>WARNING: This command is not applicable to mobile website projects. To view the complete help for this command, run `$ appbuilder help appmanager livesync`<% } %>
+
+<% if((isConsole && (isNativeScript || isCordova)) || isHtml) { %>
 ### Attributes
-`<Command>` extends the `certificate` command. You can set the following values for this attribute.
-* `create-self-signed` - Creates a self-signed certificate for code signing Android applications.
-* `remove` - Removes the selected certificate from the server.
-* `export` - Exports the selected certificate from the server on your file system.
-* `import` - Imports a certificate from your file system to the server.
+`<Platforms>` is one or more target platforms, separated by a space, for which you want to publish an update in AppManager. You can set the following target platforms.
+* `android` - Publishes an update for your Android application.
+* `ios` - Publishes an update for your iOS application.
+<% if(isCordova) { %>* `wp8` - Publishes an update for your Windows Phone application.<% } %>
+
+If you do not set a target platform, the AppBuilder CLI publishes an update for all platforms. 
+<% } %>
 
 <% if(isHtml) { %> 
+
+### Prerequisites
+
+* You must have published a major version of your app, enabled for AppManager LiveSync. To create a new major version enabled for AppManager LiveSync, run `$ appbuilder appmanager upload <Platform>`
+
+### Command Limitations
+
+* You cannot run this command on mobile website projects.
+* You cannot publish an update for Windows Phone for NativeScript projects.
+
 ### Related Commands
 
 Command | Description
@@ -23,10 +38,10 @@ Command | Description
 [appmanager upload](appmanager.html) | Allows interaction with appmanager.
 [appmanager upload android](appmanager-upload-android.html) | Builds the project and uploads the application to Telerik AppManager.
 [appmanager upload ios](appmanager-upload-ios.html) | Builds the project and uploads the application to Telerik AppManager.
-[appmanager livesync](appmanager-livesync.html) | Publish a new update of your application in Telerik AppManager.
 [appstore](appstore.html) | Allows interaction with iTunes Connect.
 [appstore list](appstore-list.html) | Lists all application records in iTunes Connect.
 [appstore upload](appstore-upload.html) | Builds the project and uploads the application to iTunes Connect.
+[certificate](certificate.html) | Lists all configured certificates for code signing iOS and Android applications with index and name.
 [certificate create-self-signed](certificate-create-self-signed.html) | Creates a self-signed certificate for code signing Android applications.
 [certificate export](certificate-export.html) | Exports a selected certificate from the server as a P12 file.
 [certificate import](certificate-import.html) | Imports an existing certificate from a P12 or a CER file stored on your local file system.
--- a/lib/commands/appmanager-livesync.ts
+++ b/lib/commands/appmanager-livesync.ts
@@ -1,0 +1,40 @@
+﻿///<reference path="../.d.ts"/>
+"use strict";
+
+export class AppManagerLiveSyncCommand implements ICommand {
+	private static ALL_PLATFORMS_OPTION = "All platforms";
+
+	constructor(private $prompter: IPrompter,
+		private $mobileHelper: Mobile.IMobileHelper,
+		private $appManagerService: IAppManagerService,
+		private $errors: IErrors,
+		private $logger:ILogger) { }
+
+	public execute(args: string[]): IFuture<void> {
+		return ((): void => {
+			if(!args || args.length === 0) {
+				var options = this.$mobileHelper.platformNames.concat(AppManagerLiveSyncCommand.ALL_PLATFORMS_OPTION);
+				var selectedPlatform = this.$prompter.promptForChoice("This command will publish a new update version to AppManager. Please select platform?", options).wait();
+				if(selectedPlatform === AppManagerLiveSyncCommand.ALL_PLATFORMS_OPTION) {
+					this.$appManagerService.publishLivePatch(this.$mobileHelper.platformNames).wait();
+				} else {
+					this.$appManagerService.publishLivePatch([selectedPlatform]).wait();
+				}
+			} else {
+				// make sure each platform is specified only once
+				var platforms = _.keys(_.groupBy(args, arg => this.$mobileHelper.normalizePlatformName(arg)));
+				this.$appManagerService.publishLivePatch(platforms).wait();
+			}
+		}).future<void>()();
+	}
+
+	public canExecute(args: string[]): IFuture<boolean> {
+		return ((): boolean => {
+			_.each(args, platform => this.$mobileHelper.validatePlatformName(platform));
+			return true;
+		}).future<boolean>()();
+	}
+
+	public allowedParameters: ICommandParameter[] = [];
+}
+$injector.registerCommand("appmanager|livesync", AppManagerLiveSyncCommand);
--- a/lib/services/appmanager-service.ts
+++ b/lib/services/appmanager-service.ts
@@ -4,8 +4,12 @@
 import constants = require("../common/mobile/constants");
 import util = require("util");
 import options = require("../options");
+import os = require("os");
+import commonOptions = require("../common/options");
 
 class AppManagerService implements IAppManagerService {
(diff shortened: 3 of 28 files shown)


lib/commands/appmanager-livesync.ts
```diff
@@ -0,0 +1,42 @@
...
+export class AppManagerLiveSyncCommand implements ICommand {
+	constructor(private $prompter: IPrompter,
+		private $mobileHelper: Mobile.IMobileHelper,
+		private $appManagerService: IAppManagerService,
+		private $errors: IErrors) { }
+
+	public execute(args: string[]): IFuture<void> {
+		return ((): void => {
+			if(!args || args.length === 0) {
+				var shouldLivePatchAllPlatforms = this.$prompter.confirm("This command will publish livepatch version to all appstores. Are you sure you want to continue?",() => true).wait();
```
ligaz: The message should be tuned a little bit :smile: Consult with @ikoevska

lib/commands/appmanager-livesync.ts
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	public canExecute(args: string[]): IFuture<boolean> {
+		return ((): boolean => {
+			var lowerCasedArguments = _.map(args, arg => arg.toLowerCase());
+
+			_.each(args, platform => this.$mobileHelper.validatePlatformName(platform));
+
+			// make sure each platform is specified only once
+			_.each(lowerCasedArguments, platform => {
```
ligaz: I can suggest another way, not sure it is better thought. You can use `_.groupBy` and print out all platforms that has count greater than 1.

rosen-vladimirov: I'm also considering `_.countBy` but I'm also not sure which of the three (current implementation, `_.countBy`, `_.groupBy`) is the best solution.

lib/services/appmanager-service.ts
```diff
@@ -48,10 +55,49 @@ class AppManagerService implements IAppManagerService {
...
+
+	public publishLivePatch(platforms: string[]): IFuture<void> {
+		return (() => {
+			platforms = _.map(platforms, platform => this.$mobileHelper.normalizePlatformName(platform));
+			var cachedOptionsRelease = commonOptions.release;
+			commonOptions.release = true;
+			this.configureLivePatchPlugin().wait();
+			var message = "If you have not published a LivePatch-enabled major version of this app before, you will not be able to distribute a LivePatch for it." + os.EOL +
+				"To create a new major version enabled for LivePatch, run the `$ appbuilder appmanager upload <Platform>` command";
+			this.$logger.warn(message);
```
ligaz: We can make the warn to accept an array of strings (along with string with union types) and make the concatenation there. This is for another PR though.

lib/services/appmanager-service.ts
```diff
@@ -48,10 +55,49 @@ class AppManagerService implements IAppManagerService {
...
+
+			this.$project.importProject().wait();
+			this.$logger.printInfoMessageOnSameLine("Publishing patch for " + platforms.join(", ") + "...");
+			this.$progressIndicator.showProgressIndicator(this.$server.tam.uploadPatch(this.$project.projectData.ProjectName, this.$project.projectData.ProjectName, <any>{ Platforms: platforms }), 2000).wait();
```
rosen-vladimirov: we must improve progressIndicator, so there will be no need to call `this.$logger.printInfoMessageOnSameLine` before and `this.$logger.printInfoMessageOnSameLine(os.EOL)` after using it. I'll handle this in a separate PR

lib/services/appmanager-service.ts
```diff
@@ -48,10 +55,49 @@ class AppManagerService implements IAppManagerService {
...
+		}).future<void>()();
+	}
+
+	public openAppManagerStore(): void {
+		var tamUrl = util.format("%s://%s/appbuilder/Services/tam", this.$config.AB_SERVER_PROTO, this.$config.AB_SERVER);
+		this.$logger.info("Go to %s to manage your apps.", tamUrl);
+		this.$opener.open(tamUrl);
+	}
+
+	private configureLivePatchPlugin(): IFuture<void> {
```
ligaz: You can move the private method bellow the public one.

lib/commands/appmanager-livesync.ts
ligaz: `_.countBy` looks like the best fit for this case.

lib/services/appmanager-service.ts
ikoevska: LivePatch instances here

docs/man_pages/publishing/appmanager-livesync.md
```diff
@@ -0,0 +1,42 @@
+appmanager livesync
+==========
+
+Usage | Syntax
+------|-------
+General | `$ appbuilder appmanager livesync [<Platforms>]`
+
+Publish new patch of the project to AppManager.
```
ikoevska: Publish a new update of your application in AppManager.

docs/man_pages/publishing/appmanager-livesync.md
```diff
@@ -0,0 +1,42 @@
...
+In case you do not specify a platform, the patch will be published for all platforms (in case you confirm the prompt dialog).
+
+`<Platforms>` can be one or more of the supported platforms: Android, iOS, WP8 (casing is not mandatory).
+
+<% if(isMobileWebsite) { %>
+This command is not applicable to Mobile Website projects.
+<% } %>
+
+<% if(isHtml) { %> 
+
```
ikoevska: ### Prerequisites

* You must have published a major version of your app, enabled for AppManager LiveSync. To create a new major version enabled for AppManager LiveSync, run \`$ appbuilder appmanager upload \<Platform\>\`

docs/man_pages/publishing/appmanager-livesync.md
```diff
@@ -0,0 +1,42 @@
...
+#### Related Commands
```
ikoevska: \#\#\# Related Commands

lib/commands/appmanager-livesync.ts
```diff
@@ -0,0 +1,43 @@
...
+	constructor(private $prompter: IPrompter,
+		private $mobileHelper: Mobile.IMobileHelper,
+		private $appManagerService: IAppManagerService,
+		private $errors: IErrors,
+		private $logger:ILogger) { }
+
+	public execute(args: string[]): IFuture<void> {
+		return ((): void => {
+			if(!args || args.length === 0) {
+				var shouldLivePatchAllPlatforms = this.$prompter.confirm("This command will publish a new patch version to AppManager. Are you sure you want to continue?",() => true).wait();
```
ikoevska: This command will publish a new update version to AppManager. Are you sure you want to continue?

lib/services/appmanager-service.ts
```diff
@@ -48,10 +55,50 @@ class AppManagerService implements IAppManagerService {
...
+
+			this.openAppManagerStore();
+		}).future<void>()();
+	}
+
+	private configureLivePatchPlugin(): IFuture<void> {
+		return (() => {
+			var plugins = this.$pluginsService.getInstalledPlugins();
+			if(!_.any(plugins, plugin => plugin.data.Identifier === AppManagerService.LIVEPATCH_PLUGIN_ID)) {
+				this.$logger.warn("AppManager LiveSync plugin is not enabled in your project. Enabling it now for release configuration.");
```
ikoevska: Same as comment for line 85

lib/services/appmanager-service.ts
```diff
@@ -48,10 +55,50 @@ class AppManagerService implements IAppManagerService {
...
+				this.$pluginsService.addPlugin(AppManagerService.LIVEPATCH_PLUGIN_ID).wait();
+				this.$logger.info("AppManager LiveSync plugin is now enabled for release configuration.");
```
ikoevska: AppManager LiveSync is now enabled for the release build configuration.

lib/commands/appmanager-livesync.ts
```diff
@@ -0,0 +1,43 @@
...
+	constructor(private $prompter: IPrompter,
+		private $mobileHelper: Mobile.IMobileHelper,
+		private $appManagerService: IAppManagerService,
+		private $errors: IErrors,
+		private $logger:ILogger) { }
+
+	public execute(args: string[]): IFuture<void> {
+		return ((): void => {
+			if(!args || args.length === 0) {
+				var shouldLivePatchAllPlatforms = this.$prompter.confirm("This command will publish a new update version to AppManager. Are you sure you want to continue?",() => true).wait();
```
Fatme: You can list all valid platforms instead of showing a confirmation message.

lib/commands/appmanager-livesync.ts
```diff
@@ -0,0 +1,43 @@
...
+				if(shouldLivePatchAllPlatforms) {
+					this.$appManagerService.publishLivePatchForAllPlatforms().wait();
+				}
+			} else {
+				this.$appManagerService.publishLivePatch(args).wait();
```
Fatme: You can remove `publishLivePatch` or `publishLivePatchForAllPlatforms` from public interface.

lib/services/appmanager-service.ts
```diff
@@ -48,10 +55,49 @@ class AppManagerService implements IAppManagerService {
...
 
-			var tamUrl = util.format("%s://%s/appbuilder/Services/tam", this.$config.AB_SERVER_PROTO, this.$config.AB_SERVER);
-			this.$logger.info("Go to %s to manage your apps.", tamUrl);
-			this.$opener.open(tamUrl);
+			this.openAppManagerStore();
+		}).future<void>()();
+	}
+
+	public openAppManagerStore(): void {
+		var tamUrl = util.format("%s://%s/appbuilder/Services/tam", this.$config.AB_SERVER_PROTO, this.$config.AB_SERVER);
```
Fatme: We have similar code in `upload` function - https://github.com/Icenium/icenium-cli/blob/master/lib/services/appmanager-service.ts#L51. I suggest to extract it to separate function and reuse the code.

ikoevska:
:+1:  for the texts

lib/services/appmanager-service.ts
teobugslayer: Also, https://github.com/nathanpeck/clui#spinnerstatustext :)

lib/commands/appmanager-livesync.ts
teobugslayer: This is inconsistent. Why does 'appmanager livesync' publish to all platforms, but 'build' errs out with requirement for explicit platform? If we find livesync's behavior to be more user-friendly, maybe in another PR we should change other commands as well?

lib/commands/appmanager-livesync.ts
```diff
@@ -0,0 +1,43 @@
...
+
+	public canExecute(args: string[]): IFuture<boolean> {
+		return ((): boolean => {
+			var result = true;
+			_.each(args, platform => this.$mobileHelper.validatePlatformName(platform));
+			
+			// make sure each platform is specified only once
+			_.each(_.countBy(args, arg => arg.toLowerCase()), (count, platform) => {
+				if(count > 1) {
+					this.$logger.error("Platform '%s' is specified %s times. Each platform should be specified only once.", platform, count.toString());
```
teobugslayer: Isn't it more user-friendly to just swallow duplicated values?

lib/services/appmanager-service.ts
```diff
@@ -48,10 +55,49 @@ class AppManagerService implements IAppManagerService {
...
+
+	public publishLivePatch(platforms: string[]): IFuture<void> {
+		return (() => {
+			platforms = _.map(platforms, platform => this.$mobileHelper.normalizePlatformName(platform));
+			var cachedOptionsRelease = commonOptions.release;
+			commonOptions.release = true;
+			this.configureLivePatchPlugin().wait();
+			var message = "If you have not published an AppManager LiveSync-enabled major version of this app before, you will not be able to distribute an AppManager LiveSync update for it." + os.EOL +
+				"To create a new major version enabled for AppManager LiveSync, run `$ appbuilder appmanager upload <Platform>`";
+			this.$logger.warn(message);
```
teobugslayer: This would print the whole message in hard-to read golden letters. Consider either printing the word "Warning" in gold, or only the first line.

lib/commands/appmanager-livesync.ts
rosen-vladimirov: I agree to change build command in another PR

docs/man_pages/publishing/appmanager-livesync.md
```diff
@@ -0,0 +1,56 @@
+appmanager livesync
+==========
+
+Usage | Syntax
```
rosen-vladimirov: Usage | Synopsis

rosen-vladimirov:
run ci

Fatme:
:+1:

PR #922, opened 2015-04-16
Throw error if the specified project directory is part of existing project

Fatme:
http://teampulse.telerik.com/view#item/290014

### DIFF
--- a/lib/commands/project/project-command-base.ts
+++ b/lib/commands/project/project-command-base.ts
@@ -1,0 +1,17 @@
+///<reference path="../../.d.ts"/>
+"use strict";
+
+export class ProjectCommandBase {
+	constructor(protected $errors: IErrors,
+		protected $project: Project.IProject) { }
+
+	protected canExecuteCore(): IFuture<boolean> {
+		return (() => {
+			if (this.$project.projectData) {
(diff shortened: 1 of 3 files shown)


lib/commands/project/project-command-base.ts
```diff
@@ -0,0 +1,12 @@
+///<reference path="../../.d.ts"/>
+"use strict";
+
+export class ProjectCommandBase {
```
rosen-vladimirov: why do we need this base class, can't we just add canExecute method of the command

teobugslayer:
:+1:

lib/commands/project/project-command-base.ts
Fatme: Maybe we can use `canExectute`, but I need to copy paste this check in `create.ts` and in `create-project.ts`.  I need it to be only on one place.

rosen-vladimirov: aaah, correct

rosen-vladimirov:
:+1:

lib/commands/project/project-command-base.ts
Fatme: Actually, this check must be in `canExecute`. I realize it after the failed QA tests :smile_cat:

PR #928, opened 2015-04-16
Use info-zip based unzip on windows, and OS supplied one on linux

teobugslayer:
and darwin


rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #929, opened 2015-04-17
Fix init command and unit tests

rosen-vladimirov:
`appbuilder init` is trying to determine project type based on the files in the current directory. As NativeScript had changed their templates we have to update our logic. Remove the check for bootstrap.js file as it is not required anymore. Fix unit tests.


Fatme:
:+1:

ligaz:
:+1:

PR #934, opened 2015-04-17
When creating a project with --path, do not create a subdir with the nam...

teobugslayer:
...e of the project

Fixes http://teampulse.telerik.com/view#item/290166


rosen-vladimirov:
:+1:

PR #936, opened 2015-04-17
Move KillSwitches dir to tmp

rosen-vladimirov:
Move KillSwitches dir to os.temp directory as when CLI is installed with sudo, we need correct privileges.

Fixes http://teampulse.telerik.com/view#item/289467
Fixes http://teampulse.telerik.com/view#item/289469
and possibly other sudo issues.


teobugslayer:
:+1:

PR #939, opened 2015-04-17
Fix no error when project name is too long

Mitko-Kerezov:
Along with some unit/integration tests

Fixes http://teampulse.telerik.com/view#item/290072
Merge after https://github.com/telerik/mobile-cli-lib/pull/292

### DIFF
--- a/lib/commands/project/create.ts
+++ b/lib/commands/project/create.ts
@@ -19,6 +19,8 @@ export class CreateCommand extends ProjectCommandBaseLib.ProjectCommandBase {
 
 	public execute(args: string[]): IFuture<void> {
 		return (() => {
+			this.validateProjectData();
+
 			var projectName = args[0];
 			var projectPath = path.join(this.$project.getNewProjectDir(), projectName);
 			this.$project.createTemplateFolder(projectPath).wait();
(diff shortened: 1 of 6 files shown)


lib/commands/project/create.ts
```diff
@@ -37,9 +37,9 @@ export class CreateCommand extends ProjectCommandBaseLib.ProjectCommandBase {
 	}
 
 	public canExecute(args: string[]): IFuture<boolean> {
-		return this.canExecuteCore();
+		return this.canExecuteCore(args[0]);
```
rosen-vladimirov: what will happen in case you write "$ appbuilder create app1 app2 app3"? This should not be allowed, but based on this canExecute method we'll use only the first parameter, instead of saying that parameters app2 and app3 are not valid for this command.

lib/commands/project/project-command-base.ts
```diff
@@ -3,15 +3,20 @@
...
 export class ProjectCommandBase {
 	constructor(protected $errors: IErrors,
+		protected $nameCommandParameter: ICommandParameter,
 		protected $project: Project.IProject) { }
 
-	protected canExecuteCore(): IFuture<boolean> {
+	protected canExecuteCore(args: string[]): IFuture<boolean> {
 		return (() => {
+			if (args.length > 1) {
+				this.$errors.failWithoutHelp("This command accepts only one argument.");
```
rosen-vladimirov: You should fail with help here. The user must understand command's syntax from the help.

Mitko-Kerezov:
run ci

lib/commands/project/project-command-base.ts
```diff
@@ -5,13 +5,11 @@ export class ProjectCommandBase {
 	constructor(protected $errors: IErrors,
 		protected $project: Project.IProject) { }
 
-	protected canExecuteCore(): IFuture<boolean> {
+	protected validateProjectData(): IFuture<boolean> {
```
rosen-vladimirov: this method returns `IFuture<boolean>`, but you never return boolean. Consider returning `IFuture<void>`

rosen-vladimirov:
:+1:  after fixing `validateProjectData` to return `IFuture<void>`
Great work!

PR #940, opened 2015-04-20
Remove obsolete autocompletion from .profile

rosen-vladimirov:
In some very old releases we were writing autocompletion information in ~/.profile. Remove this obsolete code, but do not write new autocompletion logic to this file as we do not want to do this anymore.

Fixes http://teampulse.telerik.com/view#item/289758


teobugslayer:
:+1: after green build

PR #941, opened 2015-04-20
Remove unneeded minimatch.d.ts and update to latest common lib

Fatme:
Minimatch.d.ts file was moved to common lib and we need the remove the one located in icenium-cli


rosen-vladimirov:
:+1:

teobugslayer:
:+1:

PR #942, opened 2015-04-20
Add `publish remove` to related commands in the HTML help

Mitko-Kerezov:
Related to 
http://teampulse.telerik.com/view#item/286459


teobugslayer:
:+1:

PR #944, opened 2015-04-20
Allow setting of multiple values with prop set

rosen-vladimirov:
`$ appbuilder prop set` should work for setting array values. Remove incorrect validation logic which was allowing only singe value to be used.

Fixes http://teampulse.telerik.com/view#item/289487


teobugslayer:
:+1:

PR #945, opened 2015-04-20
Fix markdown links in console

rosen-vladimirov:
Links in markdown files should be stripped in console help. Fix the incorrect regex, so now the correct text between [ and ] will be shown.
Fixes: http://teampulse.telerik.com/view#item/290105


teobugslayer:
:+1:

PR #946, opened 2015-04-20
Fix dynamic help for edit-configuration command.

rosen-vladimirov:
Mobile websites do not have configuration specific files. Executing `$ appbuilder edit-configuration -h` is trying to print list of configuration files for all frameworks. Do not print message for frameworks which do not have configuration specific files.

Fixes http://teampulse.telerik.com/view#item/289440


teobugslayer:
:+1:

PR #947, opened 2015-04-20
Show command line help for hierarchical commands

rosen-vladimirov:
If hierarchical command does NOT have default command, for example `appbuilder build`, trying to execute it should print console help. Instead this leads to opening html help.
Set option.help to true in this case. Require the options in the action itself in order to prevent cyclic errors when yok and options are required.

Fixes http://teampulse.telerik.com/view#item/290168


teobugslayer:
:+1:

PR #950, opened 2015-04-20
Fix tabtab logging

rosen-vladimirov:
Use correct `_.startsWith` method, the used one was throwing exception. Move the logic for showing `-- options` to be at the beginning of the method, so you can write `$ appbuilder --` and to see all available options. Add logic to show shortcut options. Remove OptionsService as it was required for our old logic. Now, when external options are extending knownOpts of common lib's options, we do not need this service.
Fixes http://teampulse.telerik.com/view#item/288643


teobugslayer:
:+1:

PR #951, opened 2015-04-20
Fix simulator error on Linux

rosen-vladimirov:
Move validation logic from execute method of simulate command to canExecute in order to prevent incorrect exception when command is invoked on Linux. Use hostCapabilities as top priority when validating - if the current os does not support simulator/debugger, we should fail immediately.

Fixes http://teampulse.telerik.com/view#item/284572


teobugslayer:
:+1:

PR #952, opened 2015-04-21
Fix default appid creation

rosen-vladimirov:
When user does not specify appid, we set default one based on appName. If the name starts with number, we should prefix the appid with "the" keyword. This was not working due to incorrect regex in our code. The regex was checking if the name contains only numbers. Fix this to check if the name starts with number.

Fixes http://teampulse.telerik.com/view#item/290326


Fatme:
:+1:

PR #956, opened 2015-04-21
Improve message on autocomplete fail

rosen-vladimirov:
When npm is installed with sudo, the installation command '$ sudo npm i -g appbuilder' fails to allow autocompletion as npm starts it with "nobody" user in some cases. Improve the error message in this case and show the real error only in debug mode.
Improves http://teampulse.telerik.com/view#item/288567


Fatme:
:+1:

PR #957, opened 2015-04-21
Track node versions in Analytics

teobugslayer:
Report node version to Analytics

To see the numbers, switch to the icenium-cli project insde platform analytics workspace and navigate to Feature Reports -> Feature Use -> NodeJSVersion.

Due to issues in used analytics sdk, several hacks must be made, marked with HACK comment.
The version is reported with underscores instead of points, again due to sdk limitation.

Implements http://teampulse.telerik.com/view#item/290451


rosen-vladimirov:
:+1:

PR #958, opened 2015-04-22
Fix progress indicator

Fatme:
http://teampulse.telerik.com/view#item/290449

### DIFF
--- a/lib/services/dependency-extensions-service-base.ts
+++ b/lib/services/dependency-extensions-service-base.ts
@@ -1,5 +1,6 @@
 ///<reference path="../.d.ts"/>
 "use strict";
+import os = require("os");
 import path = require("path");
 import serverExtensionsBaseLib = require("./extensions-service-base");
 import util = require("util");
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
:+1:

siyka-marinova:
:+1: :)

lib/services/dependency-extensions-service-base.ts
```diff
@@ -36,6 +37,7 @@ export class DependencyExtensionsServiceBase extends serverExtensionsBaseLib.Ext
 
 				this.$progressIndicator.showProgressIndicator(this.prepareExtensionBase(dependencyExtensionData, cachedVersion), 5000).wait();
 				this.$progressIndicator.showProgressIndicator(afterPrepareAction(), 100).wait();
+				this.$logger.printInfoMessageOnSameLine(os.EOL);
```
teobugslayer: Can we change printInfoMessageOnSameLine, so that when called without arguments, just prints os.EOL?

rosen-vladimirov: We've discussed this in personal and we'll improve all these progressIndicator and printInfo... methods in a separate PR

PR #962, opened 2015-04-23
Fix plugin remove output

rosen-vladimirov:
When you use `$appbuilder plugin remove <plugin> --debug` or `$appbuilder plugin remove <plugin> --release` you expect to remove the plugin only for the specified configuration. The output always shows that the plugin has been removed for both configurations, but in fact its removed only for the specified one.

http://teampulse.telerik.com/view#item/290590


Mitko-Kerezov:
:+1:

PR #966, opened 2015-04-23
Add check to ensure that the sb commands are executed in sb project

Fatme:


lib/services/screen-builder-service.ts
```diff
@@ -131,17 +131,35 @@ export class ScreenBuilderService implements IScreenBuilderService {
 $injector.register("screenBuilderService", ScreenBuilderService);
 
 class ScreenBuilderDynamicCommand implements ICommand {
+	private static SCREEN_BUILDER_SPECIFIC_FILES = [".yo-rc.json", ".app.json", "app.js", "tmp.app.json"];
```
Mitko-Kerezov: No *tmp.app.json* is present in a freshly created project - this would render newly created projects useless

rosen-vladimirov:
:+1:

lib/services/screen-builder-service.ts
```diff
@@ -131,17 +131,35 @@ export class ScreenBuilderService implements IScreenBuilderService {
 $injector.register("screenBuilderService", ScreenBuilderService);
 
 class ScreenBuilderDynamicCommand implements ICommand {
+	private static SCREEN_BUILDER_SPECIFIC_FILES = [".yo-rc.json", ".app.json", "app.js"];
```
ligaz: It should be just `.app.json`. That's what is used on the server to check whether or not a given project is ScreenBuilder compatible.

Mitko-Kerezov:
:+1:

PR #973, opened 2015-04-24
Fix ENOENT error in autocompletion service

rosen-vladimirov:
When we check if autocompletion or obsolete autocompletion is enabled, in case one of the shell scripts (.bash_profile for example) is missing, ENOENT error is raised. Catch it and just log it with trace option.
Fixes http://teampulse.telerik.com/view#item/290688


Mitko-Kerezov:
:+1:

PR #978, opened 2015-04-24
Fix plugin add/remove

rosen-vladimirov:
When plugin is enabled in only one configuration (for example in release only), trying to remove it with `$ appbuilder plugin remove <pluginid>` is telling you that the plugin is not enabled for this project.
When plugin is enabled in only one configuration (for example in release only) trying to add it with `$ appbuilder plugin add <pluginid>` is adding it to both configurations, so in release config the plugin is added twice.
When plugin is enabled in only one configuration (for example in release only) trying to add it with `$ appbuilder plugin add <pluginid> --debug --release` is adding it to both configurations, so in release config the plugin is added twice.

Remove `getInstalledPluginsEnabledAtLeastInOneConfiguration` and use its logic as default for getInstalledPlugins. This way we always get full list of plugins enabled in all configurations. In case you specify configuration (`--release` for example) `getInstalledPlugins` will give you only the plugins enabled in this configuration.

Fixes http://teampulse.telerik.com/view#item/290735

### DIFF
--- a/lib/services/plugins-service.ts
+++ b/lib/services/plugins-service.ts
@@ -20,7 +20,8 @@ export class PluginsService implements IPluginsService {
 		private $logger: ILogger,
 		private $project: Project.IProject,
 		private $prompter: IPrompter,
-		private $loginManager: ILoginManager) {
+		private $loginManager: ILoginManager,
+		private $projectConstants: Project.IProjectConstants) {
 
 		// Cordova plugin commands are only applicable to Cordova projects
 		this.$project.ensureCordovaProject();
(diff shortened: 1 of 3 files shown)


lib/services/plugins-service.ts
```diff
@@ -240,17 +248,18 @@ export class PluginsService implements IPluginsService {
 			if (cordovaPluginVariables && _.keys(cordovaPluginVariables[pluginData.Identifier]).length === 0) {
 				delete cordovaPluginVariables[pluginData.Identifier];
 			}
-
-			var newCorePlugins = _.without(this.$project.getProperty(PluginsService.CORE_PLUGINS_PROPERTY_NAME, configuration), plugin.toProjectDataRecord());
+			var oldCorePlugins = this.$project.getProperty(PluginsService.CORE_PLUGINS_PROPERTY_NAME, configuration);
+			var newCorePlugins = _.without(oldCorePlugins, plugin.toProjectDataRecord());
 			this.$project.setProperty(PluginsService.CORE_PLUGINS_PROPERTY_NAME, newCorePlugins, configuration);
 			this.$project.saveProject().wait();
```
Mitko-Kerezov: There is no need to save the project or set any of its properties if nothing has changes so these two lines can be moved to the if statement below

lib/services/plugins-service.ts
```diff
@@ -31,12 +31,20 @@ export class PluginsService implements IPluginsService {
 	}
 
 	public getInstalledPlugins(): IPlugin[] {
-		return this.getAllInstalledPlugins({operation: _.intersection});
-	}
+		var corePlugins: string[] = [];
+		if(options.debug) {
+			corePlugins = corePlugins.concat(this.$project.getProperty(PluginsService.CORE_PLUGINS_PROPERTY_NAME, "debug"));
```
ligaz: Don't we have constant for `"debug"`?

teobugslayer:
:+1:

Fatme:
:+1:

Mitko-Kerezov:
:+1:

tytodorov:
pr_ceco

PR #979, opened 2015-04-24
Fix indentation issues in livesync help

rosen-vladimirov:
Fix some spaces and additional symbols in livesync help, which are causing incorrect rendering of the content.


ikoevska:
:+1:

PR #981, opened 2015-04-24
Print WP8 message only for WP8 builds

rosen-vladimirov:
When you execute `$ appbuilder build <platform>` and project is using Apache Cordova 3.7.0 (or later) and WPSdk 8.0, we print warning that the user can target WPSdk 8.1
Show this warning only when building for WP8.

Fixes http://teampulse.telerik.com/view#item/290755


Fatme:
:+1:

teobugslayer:
:+1:

PR #982, opened 2015-04-27
Improve error message

Fatme:
http://teampulse.telerik.com/view#item/290832

### DIFF
--- a/lib/commands/project/create.ts
+++ b/lib/commands/project/create.ts
@@ -10,6 +10,7 @@ import ProjectCommandBaseLib = require("./project-command-base");
 
 export class CreateCommand extends ProjectCommandBaseLib.ProjectCommandBase {
 	constructor($errors: IErrors,
+		private $fs: IFileSystem,
 		private $nameCommandParameter: ICommandParameter,
 		$project: Project.IProject,
 		private $projectConstants: Project.IProjectConstants,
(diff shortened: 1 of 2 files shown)


lib/commands/project/create.ts
```diff
@@ -31,10 +32,16 @@ export class CreateCommand extends ProjectCommandBaseLib.ProjectCommandBase {
...
-			this.$screenBuilderService.prepareAndGeneratePrompt(this.$screenBuilderService.generatorName, screenBuilderOptions).wait();
-			this.$screenBuilderService.installAppDependencies().wait();
 
-			this.$project.initializeProjectFromExistingFiles(this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.Cordova).wait();
+			try {
+				this.$screenBuilderService.prepareAndGeneratePrompt(this.$screenBuilderService.generatorName, screenBuilderOptions).wait();
+				this.$screenBuilderService.installAppDependencies().wait();
+
+				this.$project.initializeProjectFromExistingFiles(this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.Cordova).wait();
+			} catch(err) {
```
teobugslayer: We can use $logger.debug(err.toString()) here

lib/project.ts
```diff
@@ -204,7 +204,7 @@ export class Project implements Project.IProject {
 			var projectDirFiles = this.$fs.readDirectory(projectDir).wait();
 
 			if(projectDirFiles.length !== 0) {
-				this.$errors.fail("The specified directory must be empty to create a new project.");
+				this.$errors.fail("The specified '%s' directory must be empty to create a new project.", projectDir);
```
teobugslayer: I am not sure, but usually the wording is ```The specified directory '%s' must be ....```. @ikoevska, your opinion?

lib/commands/project/create.ts
Fatme: Do we need this? After deleting the directory the error is thrown.

lib/project.ts
Fatme: Oo yes, it is better to be  `The specified directory '%s'`. I've changed it.

teobugslayer:
:+1:

tytodorov:
:+1:

PR #995, opened 2015-04-30
Change iOS build message

Mitko-Kerezov:
Replace the static 'Development'
with the actual provision type used for the build
Fixes http://teampulse.telerik.com/view#item/291106


rosen-vladimirov:
:+1:  after green build

PR #997, opened 2015-04-30
Fix init command

Fatme:
http://teampulse.telerik.com/view#item/291121


rosen-vladimirov:
:+1:  after green build

PR #998, opened 2015-04-30
Use latest cli-table

rosen-vladimirov:
Update cli-table version to latest tag - v.0.3.1.1
In this version there's additional fix for http://teampulse.telerik.com/view#item/291100


Fatme:
:+1:

teobugslayer:
:+1:

PR #1000, opened 2015-04-30
Provide better error handling in CLI

Mitko-Kerezov:
Stop relying on the server's message and just print the error
Addresses http://teampulse.telerik.com/view#item/290471
Depends on https://github.com/Icenium/Ice/pull/3611


teobugslayer:
:+1:

PR #1001, opened 2015-04-30
Stop execution of sample commands for on-prem deploy

teobugslayer:
Stop execution of sample command and its subcommands for on-premise deployments

Fix unit tests which breaks newly added ones

Introduce ON_PREM configuration setting
Introduce ICommand.isEnabled property with semantics:
 - when explicitly false, do not execute command
 - when true or undefined, execute it

See http://teampulse.telerik.com/view#item/288143

### DIFF
--- a/lib/commands/samples.ts
+++ b/lib/commands/samples.ts
@@ -6,22 +6,33 @@ var options: any = require("../common/options");
 
 export class PrintSamplesCommand implements ICommand {
 	constructor(private $samplesService: ISamplesService,
-		private frameworkIdentifier: string) { }
+		private frameworkIdentifier: string,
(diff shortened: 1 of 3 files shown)


lib/commands/samples.ts
```diff
@@ -6,24 +6,36 @@ var options: any = require("../common/options");
 
 export class PrintSamplesCommand implements ICommand {
 	constructor(private $samplesService: ISamplesService,
-		private frameworkIdentifier: string) { }
+		private frameworkIdentifier: string,
+		private $errors: IErrors,
```
rosen-vladimirov: errors is not required here

rosen-vladimirov:
:+1:  after green build

PR #1005, opened 2015-05-04
Encode passwords with asterisks

Mitko-Kerezov:
So that they are not outputted when --log trace option is provided
Fixes http://teampulse.telerik.com/view#item/289692
Merge after https://github.com/telerik/mobile-cli-lib/pull/318


teobugslayer:
:+1:

PR #1006, opened 2015-05-04
Fix triple import certificate attempt

Mitko-Kerezov:
When a wrong password is explicitly provided do not
attempt to import the certificate three times.
Instead prompt user for a new password
Fixes http://teampulse.telerik.com/view#item/291232


rosen-vladimirov:
:+1:  after green build

PR #1010, opened 2015-05-05
Ensure all assets are present when migrating cordova project

Mitko-Kerezov:
This way if some of them are missing, regardless whether optional or mandatory, they will be procured.

Fixes http://teampulse.telerik.com/view#item/290864


rosen-vladimirov:
:+1:

PR #1011, opened 2015-05-05
Improve update-kendoui command

rosen-vladimirov:
Improve update-kendoui command to accept several -- options:
* `--verified` - lists verified plugins
* `--core` - lists core plugins
* `--professional` - lists professional plugins

They can be combined, but core and professional cannot be used simultaneously. For example `$ appbuilder update-kendoui --core --verified` will list only verified core plugins.

Add `--latest` option. If it is specified, the user will not be prompted to select version, but the latest available version will be installed. This option can be combined with the other options. for example:
* `$ appbuilder update-kendoui --verified --latest` will install latest available verified version
* `$ appbuilder update-kendoui --verified --core --latest` will install latest available verified core version

http://teampulse.telerik.com/view#item/288592
http://teampulse.telerik.com/view#item/284131

### DIFF
--- a/lib/commands/update-kendoui.ts
+++ b/lib/commands/update-kendoui.ts
@@ -2,9 +2,14 @@
 "use strict";
 import temp = require("temp");
 import path = require("path");
+import options = require("../common/options");
 var Table = require("cli-table");
 
 class UpdateKendoUICommand implements ICommand {
+	private static VERIFIED_TAG = "verified";
+	private static KENDO_CORE = "Kendo UI Core";
+	private static KENDO_PROFESSIONAL = "Kendo UI Professional";
+
 	constructor(
 		private $errors: IErrors
 		,private $logger: ILogger
@@ -18,6 +23,20 @@ class UpdateKendoUICommand implements ICommand {
 
 	allowedParameters: ICommandParameter[] = [];
 
+	canExecute(args: string[]): IFuture<boolean> {
+		return ((): boolean => {
+			if(args && args.length > 0) {
+				this.$errors.fail("This command does not accept parameters.");
+			}
+
+			if(options.core && options.professional) {
+				this.$errors.fail("You cannot specify core and professional options simultaneously.");
+			}
+
+			return true;
+		}).future<boolean>()();
+	}
+
 	execute(args: string[]): IFuture<void> {
 		return (() => {
 			this.$loginManager.ensureLoggedIn().wait();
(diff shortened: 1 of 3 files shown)


lib/commands/update-kendoui.ts
```diff
@@ -27,32 +32,54 @@ class UpdateKendoUICommand implements ICommand {
 			}
 
 			var packages: Server.IKendoDownloadablePackageData[] = _.filter(<Server.IKendoDownloadablePackageData[]>this.$server.kendo.getPackages().wait(), p => !p.NeedPurchase);
+			if(options.verified) {
+				packages = _.filter(packages, pack => _.any(pack.VersionTags, tag => tag.toLowerCase() === UpdateKendoUICommand.VERIFIED_TAG));
+			}
+
+			if(options.core && options.professional) {
```
teobugslayer: Can this check go into canExecute callback?

lib/commands/update-kendoui.ts
```diff
@@ -62,13 +80,50 @@ class UpdateKendoUICommand implements ICommand {
...
-			var packageIdx = parseInt(choice.packageIdx, 10)-1;
-			var downloadUri = packages[packageIdx].DownloadUrl;
 			this.updateKendoFiles(downloadUri).wait();
 
 		}).future<void>()();
 	}
 
+	private selectKendoVersion(packages: Server.IKendoDownloadablePackageData[]): IFuture<string> {
+		return ((): string => {
+			this.showTableWithAvailableVersions(packages);
```
teobugslayer: Not a merge stopper, but I believe ```showAvailableVersions``` should be enough.

lib/commands/update-kendoui.ts
```diff
@@ -62,13 +80,50 @@ class UpdateKendoUICommand implements ICommand {
...
+				validate: (value: string) => {
+					var num = parseInt(value, 10);
+					return !isNaN(num) && num >= 1 && num <= packages.length ? true : "Valid values are between 1 and " + packages.length;
+				}
+			};
+
+			var choice = this.$prompter.get([schema]).wait();
+			var packageIdx = parseInt(choice.packageIdx, 10) - 1;
+			var selectedPackage = packages[packageIdx];
+			this.$logger.trace("Selected package is:");
```
teobugslayer: Can we merge the two logging lines?

teobugslayer:
:+1:  after addressing my minor comments

lib/commands/update-kendoui.ts
```diff
@@ -27,31 +46,30 @@ class UpdateKendoUICommand implements ICommand {
...
+			}
 
-			var choice = this.$prompter.get([schema]).wait();
+			if(options.professional) {
+				packages = _.filter(packages, pack => pack.Name === UpdateKendoUICommand.KENDO_PROFESSIONAL);
+			}
+
+			var downloadUri: string;
+			if(options.latest) {
+				var latestPackage = _.first(packages);
```
Fatme: It's a little bit strange to me - `var latestPackage = _.first(packages)` :smile: Is it possible packages to be empty []? If yes, `latestPackage.DownloadUrl` will throw an error.

lib/commands/update-kendoui.ts
```diff
@@ -62,13 +80,50 @@ class UpdateKendoUICommand implements ICommand {
...
 			this.updateKendoFiles(downloadUri).wait();
 
 		}).future<void>()();
 	}
 
+	private selectKendoVersion(packages: Server.IKendoDownloadablePackageData[]): IFuture<string> {
+		return ((): string => {
+			this.showTableWithAvailableVersions(packages);
+			var schema: IPromptSchema = {
+				type: "input",
```
Fatme: Can we use `list` here? :smile_cat:

lib/commands/update-kendoui.ts
rosen-vladimirov: The server returns the packages in descending order, the first package is the latest one.
If packages is empty array, it means the server has not returned any packages, which is unexpected. The other option is to specify --core and --professional (this will result in empty packages collection due to double filtering of the collection), but we've prevented this case by throwing error when both --core and --professional are used

lib/commands/update-kendoui.ts
rosen-vladimirov: I've tried this, but the list is very long and in case you want to select one of the last packages (index 70 for example), it is not very useful. So I've decided to keep current code.

lib/commands/update-kendoui.ts
rosen-vladimirov: Nope, I want to print the full object, in case I concatenate it with this string, I'll not receive the same output.

lib/commands/update-kendoui.ts
rosen-vladimirov: :+1:

teobugslayer:
:+1:

PR #1013, opened 2015-05-05
Fix error when printing errorMessage from json schemas

rosen-vladimirov:
When trying to show error from JsonSchemas and the property is not part of Base schema, an error is raised (Cannot read property 'errorMessage' of undefined).
Add check if the property is part of the schema.

Fixes http://teampulse.telerik.com/view#item/290863


Mitko-Kerezov:
:+1:

teobugslayer:
:+1:

rosen-vladimirov:
Priority of operations error, thanks @Mitko-Kerezov  for noticing this!!!
triple_facepalm

Mitko-Kerezov:
:+1:

PR #1015, opened 2015-05-06
Getting marketplace plugin identifier from MarketPlaceVersionsData

Fatme:
http://teampulse.telerik.com/view#item/286725


teobugslayer:
:+1: after green build

tytodorov:
pr_ceco after green build

PR #1018, opened 2015-05-07
var are go, let them go

teobugslayer:
a.k.a. Change "var" with "let". Fix code which uses hoisted variables.


rosen-vladimirov:
run ci

rosen-vladimirov:
:+1:  after green build

PR #1020, opened 2015-05-08
Fix error messages printing

rosen-vladimirov:
In case the property does not have error message, we do not print correct error. For example try `$ appbuilder prop set Framework Cordova` inside NativeScript project.


Mitko-Kerezov:
:+1:

PR #1025, opened 2015-05-12
Updating the system requirements for Mono

ikoevska:


rosen-vladimirov:
:+1:  after green build :)

PR #1027, opened 2015-05-13
Fix market plugins deleted upon migration

Mitko-Kerezov:
This also caused renamed plugins whose new name is a marketplace plugin to disappear
Added unit test
Fixes http://teampulse.telerik.com/view#item/291707
**Note:** The method *onWPSdkVersionChanging* has only been moved


lib/services/cordova-migration-service.ts
```diff
@@ -107,7 +107,7 @@ export class CordovaMigrationService implements ICordovaMigrationService {
 			});
 
 			let supportedPlugins = this.pluginsForVersion(toVersion).wait();
-			plugins = _.filter(plugins, plugin => _.contains(supportedPlugins, plugin));
+			plugins = _.filter(plugins, plugin => _.contains(supportedPlugins, plugin) || plugin.indexOf('@') > -1);
```
rosen-vladimirov: I suggest you to check the minimum required Cordova version of the marketplace plugin and if it is above the version, that you are going to set, print Warning. In case the min supported version is below your new Cordova version, add the plugin :)

lib/services/cordova-migration-service.ts
```diff
@@ -4,6 +4,7 @@
 import Future = require("fibers/future");
 import path = require("path");
 import helpers = require("./../helpers");
+import util = require("util");
```
rosen-vladimirov: why... :crying_cat_face:

lib/services/cordova-migration-service.ts
```diff
@@ -107,8 +117,25 @@ export class CordovaMigrationService implements ICordovaMigrationService {
 			});
 
 			let supportedPlugins = this.pluginsForVersion(toVersion).wait();
-			plugins = _.filter(plugins, plugin => _.contains(supportedPlugins, plugin));
+			plugins = _.filter(plugins, plugin => {
+				if (plugin.indexOf('@') > -1) {
+					let pluginBasicInfo = this.$pluginsService.getPluginBasicInformation(plugin);
+					if (this.isPluginSupported(pluginBasicInfo.name, pluginBasicInfo.version, toVersion)) {
+						supportedPlugins.push(plugin);
```
rosen-vladimirov: shouldn't you just return true here instead of adding the plugin to supportedPlugins and at the end check it the plugin exists in supportedPlugins?

lib/services/cordova-migration-service.ts
```diff
@@ -107,8 +117,25 @@ export class CordovaMigrationService implements ICordovaMigrationService {
...
+					} else if (!_.find(this.removedPluginsCache, p => plugin)) {
+						let remove = `Remove ${plugin} from all configuration`;
```
rosen-vladimirov: Can you move the logic from this else if in a separate method and just call it here.

lib/services/cordova-migration-service.ts
```diff
@@ -107,8 +117,25 @@ export class CordovaMigrationService implements ICordovaMigrationService {
...
+						let choice = this.$prompter.promptForChoice(`Plugin ${plugin} is not supported for Cordova version ${toVersion} what do you want to do?`, [remove, cancel]).wait();
+						if (choice === cancel) {
+							this.$errors.failWithoutHelp("Cordova migration interrupted"); 
+						} else {
+							this.removedPluginsCache.push(plugin);
+						}
+					}
+				}
 
+				return _.contains(supportedPlugins, plugin);
```
rosen-vladimirov: Shouldn't we warn the users when we remove some of the CorePlugins (not the marketplace ones)? It looks like we are silently removing the plugins.

lib/services/cordova-plugins.ts
```diff
@@ -104,10 +103,12 @@ export class CordovaPluginsService implements ICordovaPluginsService {
 	}
 
 	public getAvailablePlugins(): IFuture<Server.CordovaPluginData[]> {
+        this.$project.ensureCordovaProject();
 		return this.$server.cordova.getPlugins(this.$project.projectData.FrameworkVersion);
 	}
 
 	public createPluginData(plugin: Server.CordovaPluginData): IPlugin[] {
+        this.$project.ensureCordovaProject();
```
rosen-vladimirov: spaces instead of tabs

lib/services/cordova-plugins.ts
```diff
@@ -104,10 +103,12 @@ export class CordovaPluginsService implements ICordovaPluginsService {
 	}
 
 	public getAvailablePlugins(): IFuture<Server.CordovaPluginData[]> {
+        this.$project.ensureCordovaProject();
```
rosen-vladimirov: spaces instead of tabs

lib/services/plugins-service.ts
```diff
@@ -350,6 +373,10 @@ export class PluginsService implements IPluginsService {
 		}).future<string>()();
 	}
 
+	private normalizeMinCordovaVersion(version: string): string {
+		return _.startsWith(version, '>=') ? version.substr(2) : version;
```
rosen-vladimirov: is it possible to have version starting like `>3.0.0`? Maybe not...

lib/services/plugins-service.ts
```diff
@@ -12,23 +12,34 @@ export class PluginsService implements IPluginsService {
...
-		this.identifierToPlugin = Object.create(null);
-		Future.wait([this.createPluginsData($cordovaPluginsService),
-			this.createPluginsData($marketplacePluginsService)]);
+	private loadPluginsData(): IFuture<void> {
+		return (() => {
+			// Cordova plugin commands are only applicable to Cordova projects
+			this.$project.ensureCordovaProject();
+			this.$loginManager.ensureLoggedIn().wait();
+			this._identifierToPlugin = Object.create(null);
+			Future.wait([this.createPluginsData(this.$cordovaPluginsService),
```
teobugslayer: createPluginsData can throw, are we OK with Future.wait() swallowing the exception?

rosen-vladimirov: it will show the exceptions and break immediately after Future.wait(...)

lib/services/cordova-migration-service.ts
```diff
@@ -134,9 +161,133 @@ export class CordovaMigrationService implements ICordovaMigrationService {
...
+
+			return validValues;
+		}).future<string[]>()();
+	}
+
 	private parseMscorlibVersion(json: any): string {
 		return [json._Major, json._Minor, json._Build].join('.');
 	}
+
+	private isPluginSupported(plugin: string, version: string, migrationVersion: string): boolean {
```
rosen-vladimirov: Can you make this method public and add it to the interface. I would like to reuse it when adding new plugin to project and validate it is supported. Also I'm not sure if this is the appropriate place for this method. I think it should be part of pluginsService.

lib/services/cordova-migration-service.ts
Mitko-Kerezov: `supportedPlugins` only contains plugins which aren't from the marketplace. This is why if a valid marketplace plugin is found it is pushed to the array

lib/services/cordova-migration-service.ts
```diff
@@ -134,9 +161,133 @@ export class CordovaMigrationService implements ICordovaMigrationService {
...
+		}).future<void>()();
+	}
+
+	public getSupportedPlugins(): IFuture<string[]> {
+		return (() => {
+			let version: string;
+			if(this.$project.projectData) {
+				version = this.$project.projectData.FrameworkVersion;
+			} else {
+				let selectedFramework = _.last(_.select(this.getSupportedFrameworks().wait(), (sv: Server.FrameworkVersion) => sv.DisplayName.indexOf(this.$projectConstants.EXPERIMENTAL_TAG) === -1));
```
Fatme: Can `select` be omitted?

Mitko-Kerezov: In terms of **lodash** `select` and `filter` are aliases. 
What this code actually does is get the latest non-experimental cordova version and call `pluginsForVersion` for that version, which I'm not entirely sure is correct

lib/services/plugins-service.ts
Mitko-Kerezov: a bit of [research](http://cordova.apache.org/docs/en/3.0.0/plugin_ref_spec.md) leads me to believe that a more complex approach is indeed needed maybe we should move the [semver](https://github.com/npm/node-semver) module to common and use it here?

lib/services/plugins-service.ts
```diff
@@ -184,6 +195,19 @@ export class PluginsService implements IPluginsService {
...
+	public getPluginBasicInformation(pluginName: string): IBasicPluginInformation {
+		let result: IBasicPluginInformation = Object.create({});
+		let pluginParts = pluginName.split("@");
+		result.name = pluginParts[0];
+		result.version = pluginParts[1];
+		return result;
+	}
+
+	public isPluginSupported(plugin: string, version: string, migrationVersion: string): boolean {
+		var minCordovaVersion = _.find(this.getPluginVersions(plugin), v => v.value === version).minCordova;
```
rosen-vladimirov: use `let`

Mitko-Kerezov:
**Note:** The implementation of *onFrameworkVersionChanging* has been slightly altered!

lib/services/cordova-migration-service.ts
```diff
@@ -134,6 +143,153 @@ export class CordovaMigrationService implements ICordovaMigrationService {
...
+					this.$errors.failWithoutHelp("Unable to set %s as the target Apache Cordova version. Set the target Windows Phone SDK to 8.0 and try again.", newVersion);
+				}
+			}
+
+			let versionDisplayName = this.getDisplayNameForVersion(newVersion).wait();
+			this.$logger.info("Migrating to Cordova version %s", versionDisplayName);
+			let oldVersion = this.$project.projectData.FrameworkVersion;
+			
+			let allMarketplacePlugins: string[] = [];
+			_.each(this.$project.configurations, configuration => {
```
rosen-vladimirov: Maybe something like:
```TypeScript
let allMarketplacePlugins = _(this.$project.configurations)
						.values()
						.map(configuration => <string[]>this.$project.getProperty("CorePlugins", configuration))
						.union()
						.filter(plugin => _.contains(plugin, '@')
						.unique()
						.value();
```

lib/services/plugins-service.ts
teobugslayer: OK

lib/declarations.d.ts
```diff
@@ -417,9 +421,15 @@ interface IPlugin {
 	toProjectDataRecord(version?: string): string;
 }
 
+interface IPluginVersion {
```
teobugslayer: Consider adding JSDocs explaining the interface

lib/declarations.d.ts
```diff
@@ -405,8 +407,10 @@ interface IPluginsService {
 	printPlugins(plugins: IPlugin[]): void;
 	addPlugin(pluginName: string): IFuture<void>;
 	removePlugin(pluginName: string): IFuture<void>;
-	configurePlugin(pluginName: string, configuration?: string): IFuture<void>;
+	configurePlugin(pluginName: string, version?: string, configurations?: string[]): IFuture<void>;
```
teobugslayer: Consider adding JSDocs explaining the method

lib/services/cordova-migration-service.ts
teobugslayer: allMarketplacePlugins is actually only used once - to filter out invalid plugins. Can this operation be chained in too?

lib/services/cordova-migration-service.ts
```diff
@@ -134,6 +143,153 @@ export class CordovaMigrationService implements ICordovaMigrationService {
...
+			}
+
+			_.each(this.$project.configurations, (configuration: string) => {
+				let oldPluginsList = this.$project.getProperty("CorePlugins", configuration);
+				let newPluginsList = this.migratePlugins(oldPluginsList, oldVersion, newVersion).wait();
+				this.$logger.trace("Migrated core plugins to: ", helpers.formatListOfNames(newPluginsList, "and"));
+				this.$project.setProperty("CorePlugins", newPluginsList, configuration);
+			});
+
+			let backedUpFiles: string[] = [],
```
teobugslayer: Consider moving the code from this line bellow including the try .. finally into a dedicated function.

lib/services/plugins-service.ts
```diff
@@ -194,7 +206,21 @@ export class PluginsService implements IPluginsService {
 		}).future<void>()();
 	}
 
-	private getInstalledPluginByName(pluginName: string): IPlugin[] {
+	public getPluginBasicInformation(pluginName: string): IBasicPluginInformation {
+		let result: IBasicPluginInformation = Object.create({});
+		let pluginParts = pluginName.split("@");
+		result.name = pluginParts[0];
+		result.version = pluginParts[1];
```
teobugslayer: You can use [destructuring](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#destructuring-in-declarations-and-assignments) and collapse the 5 lines into 1.

lib/services/plugins-service.ts
```diff
@@ -345,7 +373,11 @@ export class PluginsService implements IPluginsService {
 			if (plugin.data.Name.toLowerCase() === toLowerCasePluginName || plugin.data.Identifier.toLowerCase() === toLowerCasePluginName) {
 				let pluginVersionsData = (<IMarketplacePlugin>plugin).pluginVersionsData;
 				versions = _.map(pluginVersionsData.Versions, p => {
-					return { name: p.Version, value: p.Version };
+					let pluginVersion: IPluginVersion = Object.create({});
```
teobugslayer: Again, consider using destructuring.

lib/services/cordova-migration-service.ts
```diff
@@ -134,6 +143,153 @@ export class CordovaMigrationService implements ICordovaMigrationService {
...
+
+			return this.pluginsForVersion(version).wait();
+		}).future<string[]>()();
+	}
+
+	private promptUserForInvalidPluginsAction(plugins: string[], toVersion: string): IFuture<Boolean> {
+		return (() => {
+			let remove = `Remove those plugins from all configurations`;
+			let cancel = 'Cancel Cordova migration';
+			let choice = this.$prompter.promptForChoice(`The plugins ${plugins.join(', ')} are not supported for Cordova version ${toVersion} what do you want to do?`, [remove, cancel]).wait();
```
rosen-vladimirov: in case there's only one plugin, that is not supported, the questions is:
```
? The plugins com.telerik.plugins.nativepagetransitions@0.3.0 are not supported for Cordova version 3.0.0 what do you want to do? (Use arrow keys)
```

Also do not hardcode Cordova, use this.$project.projectData.Framework instead :smile_cat:

lib/services/cordova-migration-service.ts
```diff
@@ -134,6 +143,153 @@ export class CordovaMigrationService implements ICordovaMigrationService {
...
+			if (choice === cancel) {
+				this.$errors.failWithoutHelp("Cordova migration interrupted");
+			}
+
+			return false;
```
rosen-vladimirov: you are not using this result anywhere, so consider changing the return type to IFuture<void> and omit this line

lib/declarations.d.ts
```diff
@@ -409,8 +411,17 @@ interface IPluginsService {
...
+	/**
+	 * Used to configure a plugin.
+	 * @param  {string} pluginName The name of the plugin.
+	 * @param  {string} version The version of the plugin.
+	 * @param  {string[]} configurations Configurations in which the plugin should be configured. Example: ['debug'], ['debug', 'release']
+	 * @return {IFuture<void>}
+	 */
+	configurePlugin(pluginName: string, version?: string, configurations?: string[]): IFuture<void>;
 	isPluginInstalled(pluginName: string): boolean;
+	getPluginBasicInformation(pluginName: string): IBasicPluginInformation;
```
rosen-vladimirov: it will be great to add documentation for all methods, that you've added to this interface :smiley_cat:

lib/services/cordova-migration-service.ts
```diff
@@ -134,6 +143,163 @@ export class CordovaMigrationService implements ICordovaMigrationService {
...
+				_.each(backedUpFiles, file => {
+					this.$fs.deleteFile(file + backupSuffix).wait();
+				});
+			}
+		}).future<void>()();
+	}
+
+	private promptUserForInvalidPluginsAction(plugins: string[], toVersion: string): IFuture<void> {
+		return (() => {
+			let remove = `Remove those plugins from all configurations`,
```
rosen-vladimirov: what if the plugins is only one, the questions is:
The 'plugin@version' is not supported for Cordova version 3.2.0 what do you want to do?
 - Remove those plugins from all configuraitons
...

lib/services/cordova-migration-service.ts
```diff
@@ -134,6 +143,163 @@ export class CordovaMigrationService implements ICordovaMigrationService {
...
+				cancel = 'Cancel Cordova migration',
+				pluginsString = plugins.length > 1 ? `plugins ${plugins.join(', ')} are` : `plugin ${plugins.join(', ')} is`,
+				choice = this.$prompter.promptForChoice(`The ${pluginsString} not supported for Cordova version ${toVersion} what do you want to do?`, [remove, cancel]).wait();
```
rosen-vladimirov: Maybe you need "." before what do you want to do. This will make the question more readable.

rosen-vladimirov:
Great work!!!
:+1:  after green build
And because the great refactoring needs legendary approval:
thumbsup2

Mitko-Kerezov:
run ci

PR #1028, opened 2015-05-13
Fix update-kendoui --professional --latest

rosen-vladimirov:
When user cannot use professional versions and tries to execute `appbuilder update-kendoui --professional --latest` leads to `Cannot read property 'DownloadUrl' of undefined`
The problem is that packages collection that we are using becomes empty and we are trying to get the first item of it (undefined), so it does not have DownloadUrl property.
Prevent this case by checking the length of packages collection and show correct error message in case --professional is specified.

Fixes http://teampulse.telerik.com/view#item/291742

### DIFF
--- a/lib/commands/update-kendoui.ts
+++ b/lib/commands/update-kendoui.ts
@@ -58,6 +58,14 @@ class UpdateKendoUICommand implements ICommand {
 				packages = _.filter(packages, pack => pack.Name === UpdateKendoUICommand.KENDO_PROFESSIONAL);
 			}
 
+			if(packages.length === 0) {
+				let message = "Cannot find Kendo UI packages that match the provided parameters.";
+				if(options.professional) {
+					message += " Verify that your subscription plan provides 'Kendo UI Professional'.";
+				}
+				this.$errors.failWithoutHelp(message);
+			}
+
 			let downloadUri: string;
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
ping @ikoevska  for the error messages.

lib/commands/update-kendoui.ts
```diff
@@ -58,6 +58,14 @@ class UpdateKendoUICommand implements ICommand {
 				packages = _.filter(packages, pack => pack.Name === UpdateKendoUICommand.KENDO_PROFESSIONAL);
 			}
 
+			if(packages.length === 0) {
+				let message = "There are no versions matching provided parameters.";
```
ikoevska: Cannot find Kendo UI packages that match the provided parameters.

lib/commands/update-kendoui.ts
```diff
@@ -58,6 +58,14 @@ class UpdateKendoUICommand implements ICommand {
...
+				if(options.professional) {
+					message += " Make sure your license supports 'Professional' versions of Kendo UI.";
```
ikoevska: Verify that your subscription plan provides 'Kendo UI Professional'.

teobugslayer:
:+1:

PR #1029, opened 2015-05-14
Fix upload of nativescript projects to appmanager

rosen-vladimirov:
Fix upload of nativescript projects to appmanager. The issue is raised from resolving `pluginsService` in the constructor of AppmanagerService. PluginsService constructor is failing when the project is **NOT** Cordova.
Use injector to resolve pluginsService only where it is required.
Fix error when plugin is enabled manually in configuration, but we are unable to find it in marketplace (for example manually add com.telerik.LivePatch in release config and try $ appbuilder appmanager livesync android with user who does not have access to this feature. You'll receive error **Cannot read property data of undefined**

Fixes http://teampulse.telerik.com/view#item/291755

### DIFF
--- a/lib/services/appmanager-service.ts
+++ b/lib/services/appmanager-service.ts
@@ -19,10 +19,10 @@ class AppManagerService implements IAppManagerService {
 		private $loginManager: ILoginManager,
 		private $opener: IOpener,
 		private $buildService: Project.IBuildService,
-		private $pluginsService: IPluginsService,
 		private $progressIndicator: IProgressIndicator,
 		private $mobileHelper: Mobile.IMobileHelper,
-		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants) { }
+		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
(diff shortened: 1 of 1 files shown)


teobugslayer:
:+1:

tytodorov:
pr_ceco
After green build

Fatme:
:+1:

lib/services/appmanager-service.ts
```diff
@@ -71,8 +70,8 @@ class AppManagerService implements IAppManagerService {
 			this.$loginManager.ensureLoggedIn().wait();
 
 			platforms = _.map(platforms, platform => this.$mobileHelper.normalizePlatformName(platform));
-			var cachedOptionsRelease = commonOptions.release;
-			commonOptions.release = true;
+			var cachedOptionsRelease = options.release;
```
teobugslayer: This code is conceptually wrong. It may be left for the hotfix release but it must be rewritten correctly in master.

rosen-vladimirov: I'll fix this in master branch

PR #1034, opened 2015-05-15
Compile Yok in strict mode

teobugslayer:


rosen-vladimirov:
:+1:  after green build

PR #1036, opened 2015-05-15
Detect mono version

teobugslayer:
At install, warn if there is problem with mono installation

Implements http://teampulse.telerik.com/view#item/291901


rosen-vladimirov:
:+1:

teobugslayer:
This has already changed in another PR

PR #1038, opened 2015-05-15
Fixes

teobugslayer:
* Fix incorrect validation for iTunes on Linux - we consider it existing.
 * Fix console not released when only 1 CLI instance is running


rosen-vladimirov:
:+1:

PR #1041, opened 2015-05-19
Use access_token for samples

rosen-vladimirov:
When we call api.github.com, there are some limitations. In order to bypass them, we have to pass access token to the call. Take it from ~/.abgithub file and add it to our "samples" calls.
More info here http://teampulse.telerik.com/view#item/291726

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -6,6 +6,7 @@ import fiber = require("fibers");
 import helpers = require("./helpers");
 import staticConfigBaseLib = require("./common/static-config-base");
 import configBaseLib = require("./common/config-base");
+import osenv = require("osenv");
 
 export class Configuration extends configBaseLib.ConfigBase implements IConfiguration { // User specific config
 	AB_SERVER_PROTO: string;
@@ -81,6 +82,7 @@ export class Configuration extends configBaseLib.ConfigBase implements IConfigur
 $injector.register("config", Configuration);
 
 export class StaticConfig extends staticConfigBaseLib.StaticConfigBase implements IStaticConfig {
+	private static TOKEN_FILENAME = ".abgithub";
 	public PROJECT_FILE_NAME = ".abproject";
 	public CLIENT_NAME = "AppBuilder";
 	public ANALYTICS_API_KEY = "13eaa7db90224aa1861937fc71863ab8";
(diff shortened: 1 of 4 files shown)


lib/declarations.d.ts
```diff
@@ -230,6 +230,7 @@ interface IConfiguration extends Config.IConfig {
 }
 
 interface IStaticConfig extends Config.IStaticConfig {
+	GITHUB_ACCESS_TOKEN_FILEPATH: string;
```
teobugslayer: JSDoc FTW!

lib/services/samples-service.ts
```diff
@@ -82,8 +82,8 @@ export class SamplesService implements ISamplesService {
 				let filepath = path.join(tempDir, sampleName);
 				let file = this.$fs.createWriteStream(filepath);
 				let fileEnd = this.$fs.futureFromEvent(file, "finish");
-
-				let response = this.$httpClient.httpRequest({ url: sample.zipUrl, pipeTo: file }).wait();
+				let accessToken = this.getGitHubAccessToken("?").wait();
+				let response = this.$httpClient.httpRequest({ url: `${sample.zipUrl}${accessToken}`, pipeTo: file }).wait();
```
teobugslayer: Good, good! Let the ${}s flow through!

lib/config.ts
```diff
@@ -95,6 +96,10 @@ export class StaticConfig extends staticConfigBaseLib.StaticConfigBase implement
 	public SYS_REQUIREMENTS_LINK = "http://docs.telerik.com/platform/appbuilder/running-appbuilder/running-the-cli/system-requirements-cli";
 	public SOLUTION_SPACE_NAME = "Private_Build_Folder";
 	public QR_SIZE = 300;
+	public get GITHUB_ACCESS_TOKEN_FILEPATH(): string {
+		let tokenFileName = ".abgithub";
```
teobugslayer: Shouldn't this be private static?

lib/services/samples-service.ts
```diff
@@ -218,5 +219,26 @@ export class SamplesService implements ISamplesService {
 		let matchedCategory = _.find(sortedCategories, category => category.regEx.test(description));
 		return matchedCategory ? matchedCategory.id : null;
 	}
+	
+	private getGitHubAccessToken(queryToken: string): IFuture<string> {
```
teobugslayer: This function is badly named. It does not return the token. It returns a part of an url, containing the token.

teobugslayer:
:+1:

PR #1046, opened 2015-05-20
Fix plugin configure

rosen-vladimirov:
When plugin configure is called for plugin, which is not installed, it automatically adds the plugin to your project. We already have the validation implemented, but after one refactoring, the isPluginInstalled method was returning incorrect result.
Added unit tests for isPluginInstalled method.

Update to latest common lib where definitions of semver had been updated.


teobugslayer:
:+1:

PR #1049, opened 2015-05-22
Allow migration of NativeScript projects

rosen-vladimirov:
When the project is migrated we have to update tns_modules directory and .abproject with the new version of the project. Add new NativeScriptMigrationService and use it with dynamic call to upgrade the project. In case an error occurs during migration, rollback the migration. In case current project version is not supported, we have to break the migration. Oboslete versions can still be build, but we cannot migrate them.
Download NativeScript migration data during prepublish. Make sure to save supportedVersion in format - DisplayName(string) and Version(string).
You can change the version by using: `$ appbuilder prop set FrameworkVersion 1.0.1` (or 0.10.0) or `$ appbuilder mobileFramework set 1.0.1`
You can list valid versions by using `$ appbuilder prop print FramworkVersion --validValue` or `$ appbuilder mobileFramework`

Implements http://teampulse.telerik.com/view#item/292172

### DIFF
--- a/lib/commands/dev/prepackage.ts
+++ b/lib/commands/dev/prepackage.ts
@@ -4,12 +4,13 @@
 import Future = require("fibers/future");
 
 export class PrePackageCommand implements ICommand {
-	constructor(private $cordovaMigrationService: ICordovaMigrationService,
+	constructor(private $cordovaMigrationService: IFrameworkMigrationService,
 		private $jsonSchemaLoader: IJsonSchemaLoader,
 		private $logger: ILogger,
 		private $resourceDownloader: IResourceDownloader,
 		private $serviceProxy: Server.IServiceProxy,
-		private $templatesService: ITemplatesService) { }
+		private $templatesService: ITemplatesService,
+		private $nativeScriptMigrationService: IFrameworkMigrationService) { }
 
 	public disableAnalytics = true;
 
@@ -28,11 +29,12 @@ export class PrePackageCommand implements ICommand {
 			this.$logger.info("Unpacking app resources.");
 			this.$templatesService.unpackAppResources().wait();
 			this.$logger.info("Downloading cordova migration data.");
-			this.$cordovaMigrationService.downloadCordovaMigrationData().wait();
-			//Cordova files have to be downloaded after cordova migration data so we know which cordova versions we support
+			this.$cordovaMigrationService.downloadMigrationData().wait();
+			// Cordova files have to be downloaded after cordova migration data so we know which cordova versions we support
 			this.$logger.info("Downloading cordova.js files.");
 			this.$resourceDownloader.downloadCordovaJsFiles().wait();
-
+			this.$logger.info("Downloading nativescript migration data.")
+			this.$nativeScriptMigrationService.downloadMigrationData().wait();
 			this.$serviceProxy.setShouldAuthenticate(true);
 		}).future<void>()();
 	}
--- a/lib/commands/dev/prepackage.ts
+++ b/lib/commands/dev/prepackage.ts
@@ -28,12 +28,12 @@ export class PrePackageCommand implements ICommand {
 			this.$jsonSchemaLoader.downloadSchemas().wait();
 			this.$logger.info("Unpacking app resources.");
 			this.$templatesService.unpackAppResources().wait();
-			this.$logger.info("Downloading cordova migration data.");
+			this.$logger.info("Downloading Cordova migration data.");
 			this.$cordovaMigrationService.downloadMigrationData().wait();
 			// Cordova files have to be downloaded after cordova migration data so we know which cordova versions we support
 			this.$logger.info("Downloading cordova.js files.");
 			this.$resourceDownloader.downloadCordovaJsFiles().wait();
-			this.$logger.info("Downloading nativescript migration data.")
+			this.$logger.info("Downloading NativeScript migration data.")
 			this.$nativeScriptMigrationService.downloadMigrationData().wait();
 			this.$serviceProxy.setShouldAuthenticate(true);
 		}).future<void>()();
--- a/lib/commands/framework-versions/print-versions.ts
+++ b/lib/commands/framework-versions/print-versions.ts
@@ -2,18 +2,19 @@
 "use strict";
 
 export class PrintFrameworkVersionsCommand implements ICommand {
-	constructor(private $cordovaMigrationService: ICordovaMigrationService,
+	constructor(private $cordovaMigrationService: IFrameworkMigrationService,
+		private $nativeScriptMigrationService: IFrameworkMigrationService,
 		private $project: Project.IProject,
 		private $logger: ILogger,
-		private $errors: IErrors) { }
+		private $errors: IErrors,
+		private $projectConstants: Project.IProjectConstants) { }
 
 	public execute(args: string[]): IFuture<void> {
 		return (() => {
-			let supportedVersions: Server.FrameworkVersion[] = this.$cordovaMigrationService.getSupportedFrameworks().wait();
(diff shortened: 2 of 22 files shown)


lib/commands/dev/prepackage.ts
```diff
@@ -28,11 +29,12 @@ export class PrePackageCommand implements ICommand {
...
 			this.$logger.info("Downloading cordova migration data.");
-			this.$cordovaMigrationService.downloadCordovaMigrationData().wait();
-			//Cordova files have to be downloaded after cordova migration data so we know which cordova versions we support
+			this.$cordovaMigrationService.downloadMigrationData().wait();
+			// Cordova files have to be downloaded after cordova migration data so we know which cordova versions we support
 			this.$logger.info("Downloading cordova.js files.");
 			this.$resourceDownloader.downloadCordovaJsFiles().wait();
-
+			this.$logger.info("Downloading nativescript migration data.")
+			this.$nativeScriptMigrationService.downloadMigrationData().wait();
```
Fatme: Do we need to download ns migration data here? Can we do it when we need it?

rosen-vladimirov: Well it's just 5 mb and with NativeScript 1.1 we will not have to do it, so I've decided to keep current processes and have the migration data as part of our package.

lib/commands/framework-versions/print-versions.ts
```diff
@@ -26,8 +27,10 @@ export class PrintFrameworkVersionsCommand implements ICommand {
 
 	public canExecute(args: string[]): IFuture<boolean> {
 		return (() => {
-			this.$project.ensureCordovaProject();
-
+			this.$project.ensureProject();
+			if(this.$project.projectData.Framework !== this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.Cordova && this.$project.projectData.Framework !== this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.NativeScript) {
```
Fatme: Can we add new property to `projectCapabilities` ? Actually only `website` project should throw error, `cordova` and `nativescript` should return true.

lib/commands/framework-versions/print-versions.ts
```diff
@@ -2,18 +2,19 @@
...
 		private $project: Project.IProject,
 		private $logger: ILogger,
-		private $errors: IErrors) { }
+		private $errors: IErrors,
+		private $projectConstants: Project.IProjectConstants) { }
 
 	public execute(args: string[]): IFuture<void> {
 		return (() => {
-			let supportedVersions: Server.FrameworkVersion[] = this.$cordovaMigrationService.getSupportedFrameworks().wait();
+			let migrationService = this.$project.projectData.Framework === this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.Cordova ? this.$cordovaMigrationService : this.$nativeScriptMigrationService;
```
Fatme: Consider exposing `migrationService` in `project`. This way you can use - `this.$project.migrationService.getSupportedFrameworks`.

rosen-vladimirov: I've considered this, but we have some cyclic dependencies with this case. Project will resove cordova project, which will resolve cordova migration service which will resolve plugins service, which will try to resolve project
I have an idea how to fix this, but I'll handle it in a separate PR (resolving all these cyclic dependencies).

lib/declarations.d.ts
```diff
@@ -384,11 +384,11 @@ interface IPathFilteringService {
...
 }
 
-interface ICordovaMigrationService {
-	downloadCordovaMigrationData(): IFuture<void>;
+interface IFrameworkMigrationService {
+	downloadMigrationData(): IFuture<void>;
 	getSupportedVersions(): IFuture<string[]>;
-	pluginsForVersion(version: string): IFuture<string[]>;
-	migratePlugins(plugins: string[], fromVersion: string, toVersion: string): IFuture<string[]>;
+	pluginsForVersion?(version: string): IFuture<string[]>;
```
Fatme: Currently `plugins` is valid only for cordova but this function is part from `IFrameworkMigrationService` so I suggest to rename it to `migrationDataForVersion`

lib/project.ts
```diff
@@ -630,6 +630,18 @@ export class Project implements Project.IProject {
 		}).future<void>()();
 	}
 
+	public isTypeScriptProject(): IFuture<boolean> {
```
Fatme: I think we have similar function in `typescript-compilation-service`. Can we use it ?

lib/services/nativescript-migration-service.ts
```diff
@@ -0,0 +1,143 @@
...
+"use strict";
+
+import Future = require("fibers/future");
+import path = require("path");
+import helpers = require("./../helpers");
+
+export class NativeScriptMigrationService implements IFrameworkMigrationService {
+	private static TYPESCRIPT_ABBREVIATION = "TS";
+	private static JAVASCRIPT_ABBREVIATION = "JS";	
+	private nativeScriptMigrationFile: string = path.join(__dirname, "../../resources/NativeScript", "nativeScript-migration-data.json");
```
Fatme: You can use `this.$resources.resolvePath("NativeScript")` - this will return the path to NativeScript folder

lib/services/nativescript-migration-service.ts
```diff
@@ -0,0 +1,143 @@
...
+			let supportedVersions = _.map(json.SupportedVersions, supportedVersion => {
+				supportedVersion.Version = this.parseMscorlibVersion(supportedVersion.Version);
+				return supportedVersion;
+			});
+			_.each(json.ObsoleteVersions, obsoleteVersion => {
+				obsoleteVersion.Version = this.parseMscorlibVersion(obsoleteVersion.Version); 
+			});
+
+			this.$fs.writeJson(this.nativeScriptMigrationFile, json).wait();
+			let supportedLanguages = [NativeScriptMigrationService.JAVASCRIPT_ABBREVIATION, NativeScriptMigrationService.TYPESCRIPT_ABBREVIATION];
```
Fatme: This can be static array

lib/commands/framework-versions/print-versions.ts
Fatme: :cool:  :+1:

lib/commands/framework-versions/print-versions.ts
rosen-vladimirov: I like the idea, I'll do it :)

lib/commands/dev/prepackage.ts
```diff
@@ -28,11 +29,12 @@ export class PrePackageCommand implements ICommand {
...
 			this.$templatesService.unpackAppResources().wait();
 			this.$logger.info("Downloading cordova migration data.");
-			this.$cordovaMigrationService.downloadCordovaMigrationData().wait();
-			//Cordova files have to be downloaded after cordova migration data so we know which cordova versions we support
+			this.$cordovaMigrationService.downloadMigrationData().wait();
+			// Cordova files have to be downloaded after cordova migration data so we know which cordova versions we support
 			this.$logger.info("Downloading cordova.js files.");
 			this.$resourceDownloader.downloadCordovaJsFiles().wait();
-
+			this.$logger.info("Downloading nativescript migration data.")
```
teobugslayer: Shouldn't it be spelled "NativeScript"?

lib/commands/framework-versions/print-versions.ts
teobugslayer: good idea!

lib/declarations.d.ts
rosen-vladimirov: In fact I'll extact another interface for CordovaMigration, that will extend this one. This way I'll keep the code cleaner

lib/project.ts
rosen-vladimirov: It is in our hooks - typescript-compilation.ts
I'll reuse it there and keep the code here.

PR #1054, opened 2015-05-25
Add latest or default plugin versions

rosen-vladimirov:
Add support for installing the latest plugin version. The command is: `$ appbuilder plugin add toast --latest`. This will install the latest available version of Toast plugin.
Add support for installing the default plugin version. The command is: `$ appbuilder plugin add toast --default`. This will install the default version of Toast plugin.
As on the server MarketplacePluginVersionsData has three more properties, which are not generated in our server-api, implement IMarketplacePluginVersionsData interface, that extends this class and add the properties there.

http://teampulse.telerik.com/view#item/292306

### DIFF
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -501,8 +501,26 @@ interface IBasicPluginInformation {
 	version: string;
 }
 
+/**
+ * Extends Server's MarketplacePluginVersionsData interface.
+ */
+interface IMarketplacePluginVersionsData extends Server.MarketplacePluginVersionsData {
+	/**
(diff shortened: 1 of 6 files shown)


lib/declarations.d.ts
```diff
@@ -501,8 +501,14 @@ interface IBasicPluginInformation {
 	version: string;
 }
 
+interface IMarketplacePluginVersionsData extends Server.MarketplacePluginVersionsData {
```
teobugslayer: consider adding some JSDocs

teobugslayer:
:+1: after addressing my small comment

PR #1057, opened 2015-05-26
Allow migration of NativeScript projects

rosen-vladimirov:
When the project is migrated we have to update tns_modules directory and .abproject with the new version of the project. Add new NativeScriptMigrationService and use it with dynamic call to upgrade the project. In case an error occurs during migration, rollback the migration. In case current project version is not supported, we have to break the migration. Oboslete versions can still be build, but we cannot migrate them.
Download NativeScript migration data during prepublish. Make sure to save supportedVersion in format - DisplayName(string) and Version(string).
You can change the version by using: `$ appbuilder prop set FrameworkVersion 1.0.1` (or 0.10.0) or `$ appbuilder mobileFramework set 1.0.1`
You can list valid versions by using `$ appbuilder prop print FramworkVersion --validValue` or `$ appbuilder mobileFramework`

Add two new interfaces, one for NativeScriptMigration data and one describing Framework versions (with version and display name). They second one is used to unify the way we get supported framework versions in Cordova and NativeScript projects. Until now, we were calling the server to get the Cordova supported framework versions. In fact we can have this information stored in cordova-migration-data.json. Store it there and when we want to check the data, read the file. The same behavior is used for NativeScript.

Implements http://teampulse.telerik.com/view#item/292172

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ npm-debug.log
 node_modules
 resources/App_Resources
 resources/Cordova
+resources/NativeScript
 resources/ItemTemplates
 resources/ProjectTemplates
 resources/json-schemas
--- a/docs/man_pages/project/configuration/mobileframework-set.md
+++ b/docs/man_pages/project/configuration/mobileframework-set.md
@@ -5,17 +5,16 @@ Usage | Synopsis
 ------|-------
 General | `$ appbuilder mobileframework set <Version> [--path <Directory>]`
 
-Sets the selected Apache Cordova version for the project and updates the enabled core or integrated plugins to match it.
+<% if(isHtml) { %>Sets the selected framework version for the project.<% } %>
 
 <% if(isConsole)  { %>
-<% if(isNativeScript)  { %>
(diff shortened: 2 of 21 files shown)


teobugslayer:
Executing
```bash
$ ../../bin/appbuilder prop set FrameworkVersion 1.0.1
1.0.1
```

I think this UX can be improved. Can we print something similar to: "Framework version changed to 1.0.1"?

teobugslayer:
Aside from my small remark, :+1:

teobugslayer:
Noticed that:
```bash
$ ../../bin/appbuilder prop set FrameworkVersion 1.0.1
? You cannot build with the Windows Phone 8.1 SDK with the currently selected target version of Apache Cordova. Do you want to switch to Windows Phone 8.0 SDK? Yes
Migrating to WPSdk version 8.0
Cannot find version 1.0.1 in the supported versions.
```

rosen-vladimirov:
I've handled the issue with Windows Phone 8.1 question. The other one will be handled in a separate PR.

PR #1058, opened 2015-05-26
Set Description when creating new project

rosen-vladimirov:
When a new Cordova project is created, we have to set its Description to be equal to projectName. This is due to this change: https://github.com/Icenium/Ice/pull/3729
This will fix our unit tests.

> NOTE: The change on the server reflects only Cordova projects, but it will be changed to update NativeScript project's descriptions as well.


teobugslayer:
:+1: after green build

Fatme:
:+1:

PR #1062, opened 2015-05-27
Update sb to 0.0.7

Fatme:
http://teampulse.telerik.com/view#item/291564

### DIFF
--- a/lib/services/app-scaffolding-extensions-service.ts
+++ b/lib/services/app-scaffolding-extensions-service.ts
@@ -5,6 +5,7 @@ import dependencyExtensionsServiceLib = require("./dependency-extensions-service
 import hostInfo = require("./../common/host-info");
 import path = require("path");
 import util = require("util");
+import Future = require("fibers/future");
 
 export class AppScaffoldingExtensionsService extends dependencyExtensionsServiceLib.DependencyExtensionsServiceBase implements IAppScaffoldingExtensionsService {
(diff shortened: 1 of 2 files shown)


lib/services/app-scaffolding-extensions-service.ts
```diff
@@ -30,6 +30,12 @@ export class AppScaffoldingExtensionsService extends dependencyExtensionsService
 			let afterPrepareAction = () => {
 				return (() => {
 					this.npmInstall("glob-watcher@0.0.8").wait(); // HACK: With this we are able to make paths shorter with 20 symbols.
+					
+					let generatorBaseDependencies = require(path.join(this.appScaffoldingPath, "node_modules", "screen-builder-base-generator", "package.json")).dependencies;
+					_.each(generatorBaseDependencies, (value, key) => {
```
rosen-vladimirov: You can simplify this by using:
```TypeScript
_.each(generatorBaseDependencies, (value, key) => this.npmInstall(`${key}@${value}`).wait());
```
Also we can speed it up by using different futures for each install, for example
```TypeScript
Future.wait(_.map(generatorBaseDependencies, (value, key) => this.npmInstall(`${key}@${value}`));
```
> NOTE: I've not tested this Future.wait(...) code here

rosen-vladimirov:
:+1:  after using Futures :)

PR #1063, opened 2015-05-27
Show only supported versions for Cordova

rosen-vladimirov:
`appbuilder mobileframework` lists the supported versions of Cordova framework. Currently our server supports Cordova 2.7.0 (marked as deprecated), but CLI requires at least 3.0.0. Executing the command `appbuilder mobileframework` insider Cordova project lists version 2.7.0. The problem is in the Cordova migration data file, where we are not filtering versions below 3.0.0. Fix this by filtering the versions and add save in the file only the supported versions.

### DIFF
--- a/lib/services/cordova-migration-service.ts
+++ b/lib/services/cordova-migration-service.ts
@@ -124,7 +124,10 @@ export class CordovaMigrationService implements ICordovaMigrationService {
 			_.each(cliSupportedVersions, version => {
 				integratedPlugins[version] = json.IntegratedPlugins[version];
 			});
-			let supportedFrameworkVersion: IFrameworkVersion[] = _.map(json.SupportedFrameworkVersions, fv => { return {displayName: fv.DisplayName, version: this.parseMscorlibVersion(fv.Version)} });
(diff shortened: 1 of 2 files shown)


Fatme:
:+1:

lib/services/cordova-migration-service.ts
```diff
@@ -124,7 +124,10 @@ export class CordovaMigrationService implements ICordovaMigrationService {
 			_.each(cliSupportedVersions, version => {
 				integratedPlugins[version] = json.IntegratedPlugins[version];
 			});
-			let supportedFrameworkVersion: IFrameworkVersion[] = _.map(json.SupportedFrameworkVersions, fv => { return {displayName: fv.DisplayName, version: this.parseMscorlibVersion(fv.Version)} });
+			let supportedFrameworkVersion: IFrameworkVersion[] = _(json.SupportedFrameworkVersions)
```
ligaz: It is sad that we are using `_` when `map` and `filter` are built-in. :cry:

rosen-vladimirov: We love `lodash` (and keep the code consistent in our codebase)

PR #1064, opened 2015-05-28
Fix iTunes validator on windows

Fatme:
http://teampulse.telerik.com/view#item/292423


rosen-vladimirov:
:+1:  after green/ reasonably yellow build

PR #1065, opened 2015-05-28
Use AdHoc provision first if possible

Mitko-Kerezov:
If a **valid** AdHoc provision is present it should be chosen with priority over the others (due to the fact that it can generate a QR code)
Fixes http://teampulse.telerik.com/view#item/292411

### DIFF
--- a/lib/services/build.ts
+++ b/lib/services/build.ts
@@ -176,7 +176,17 @@ export class BuildService implements Project.IBuildService {
 					}
 				} else if(!settings.buildForiOSSimulator) {
 					let deviceIdentifier = settings.device ? settings.device.getIdentifier() : undefined;
-					provisionData = this.$identityManager.autoselectProvision(appIdentifier, settings.provisionTypes, deviceIdentifier).wait();
+					try {
+						provisionData = this.$identityManager.autoselectProvision(appIdentifier, [constants.ProvisionType.AdHoc], deviceIdentifier).wait();
+					} catch (error) {
+						this.$logger.warn("Cannot generate QR code because an applicable AdHoc provisioning profile is not available.");
+						let additionalInfo = error.message.split(os.EOL)[1];
+						if (additionalInfo) {
+							this.$logger.warn(additionalInfo);
+						}
+
+						provisionData = this.$identityManager.autoselectProvision(appIdentifier, _.values(constants.ProvisionType), deviceIdentifier).wait();
+					}
 					this.$options.provision = provisionData.Name;
 				}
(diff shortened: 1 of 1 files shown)


lib/services/build.ts
```diff
@@ -176,7 +176,17 @@ export class BuildService implements Project.IBuildService {
 					}
 				} else if(!settings.buildForiOSSimulator) {
 					let deviceIdentifier = settings.device ? settings.device.getIdentifier() : undefined;
-					provisionData = this.$identityManager.autoselectProvision(appIdentifier, settings.provisionTypes, deviceIdentifier).wait();
+					try {
+						provisionData = this.$identityManager.autoselectProvision(appIdentifier, [constants.ProvisionType.AdHoc], deviceIdentifier).wait();
+					} catch (error) {
+						this.$logger.warn("Cannot find suitable AdHoc provision so no QR code will be generated");
```
teobugslayer: "Cannot generate QR code because an applicable AdHoc provisioning profile is not available.", as @ikoevska suggested.

teobugslayer:
:+1: after changing the message.

Mitko-Kerezov:
run ci

Mitko-Kerezov:
run ci

Mitko-Kerezov:
run ci

lib/services/build.ts
```diff
@@ -176,7 +176,17 @@ export class BuildService implements Project.IBuildService {
...
 				} else if(!settings.buildForiOSSimulator) {
 					let deviceIdentifier = settings.device ? settings.device.getIdentifier() : undefined;
-					provisionData = this.$identityManager.autoselectProvision(appIdentifier, settings.provisionTypes, deviceIdentifier).wait();
+					try {
+						provisionData = this.$identityManager.autoselectProvision(appIdentifier, [constants.ProvisionType.AdHoc], deviceIdentifier).wait();
+					} catch (error) {
+						this.$logger.warn("Cannot generate QR code because an applicable AdHoc provisioning profile is not available.");
+						let additionalInfo = error.message.split(os.EOL)[1];
+						if (additionalInfo) {
+							this.$logger.warn(additionalInfo);
```
rosen-vladimirov: why not trace?

Mitko-Kerezov: because this may be valuable information to the user like for example
Cannot use provision X because no valid certificates found
Which would be a reminder that the user imported the provision only and forgot about certificates

rosen-vladimirov:
:+1:

Mitko-Kerezov:
run ci

Mitko-Kerezov:
run ci

PR #1068, opened 2015-05-28
Update CorePlugins in all config files

rosen-vladimirov:
CorePlugins could be part of .debug.abproject, .release.abproject and .abproject. Make sure `prop` related commands will do the same as our UI Clients:
* When CorePlugins in debug and release are the same, move them to .abproject.
* When a plugin is added/removed for only one configuration, move all plugins to configuration specific files and modify the one, specified by user.
* When prop set is used without configuration flag (--debug or --release), remove the CorePlugins from configuration specific files and set them in .abproject.
* When Core Plugins are different in each configuration and user had not specified configuration flag for operation, execute it on all available configs.
* When different plugins are available in .abproject and any configuration specific file, throw error as we are not sure which of the operations to execute.

Add unit tests for this behavior.
http://teampulse.telerik.com/view#item/285810

### DIFF
--- a/lib/services/project-properties-service.ts
+++ b/lib/services/project-properties-service.ts
@@ -61,22 +61,20 @@ export class ProjectPropertiesService implements IProjectPropertiesService {
 		return dataToBeUpdated;
 	}
 
-	private validateProjectData(projectData: IProjectData, configurationSpecificData?: IProjectData): void {
-		let dataToValidate = Object.create(null);
-		_.extend(dataToValidate, projectData);
-		if(configurationSpecificData) {
-			_.extend(dataToValidate, configurationSpecificData);
-		}
-		this.$jsonSchemaValidator.validate(dataToValidate);
-	}
-
-	private notifyPropertyChanged(framework: string, propertyName: string, propertyValue: any): IFuture<void> {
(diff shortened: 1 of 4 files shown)


Mitko-Kerezov:
My :+1: and a job very well done!

lib/services/project-properties-service.ts
```diff
@@ -259,5 +257,78 @@ export class ProjectPropertiesService implements IProjectPropertiesService {
...
+	private validateAllProjectData(projectData: IProjectData, configurationSpecificData:IDictionary<IProjectData>): void {
+		let projectConfigurations = _.keys(configurationSpecificData);
+		_.each(projectConfigurations, configuration => {
+			this.validateProjectData(projectData, configurationSpecificData[configuration]);
+		});
+
+		this.validateProjectData(projectData);
+	}
+
+	private validateProjectData(projectData: IProjectData, configurationSpecificData?: IProjectData): void {
```
Fatme: We have some validation logic in `project.ts`. Can we unify it and use only from one place?

rosen-vladimirov:
Closing this PR and reopening it for release branch.

PR #1069, opened 2015-05-28
Detect Core plugins using the RegEx in cordova.json

Fatme:
http://teampulse.telerik.com/view#item/292500


rosen-vladimirov:
:+1:

PR #1073, opened 2015-05-29
Fix debugging of NativeScript applications

Fatme:
http://teampulse.telerik.com/view#item/292771

Should be merged after this PR https://github.com/Icenium/Ice/pull/3762


teobugslayer:
:+1:

PR #1074, opened 2015-05-29
Fix warning message when building with --download

Mitko-Kerezov:
When building for iOS with `--download` a QR code cannot be generated so there's no point in a warning message
Fixes http://teampulse.telerik.com/view#item/292777


Fatme:
:+1:

Mitko-Kerezov:
run ci

PR #1076, opened 2015-05-29
Add LivePatch to available plugins

rosen-vladimirov:
Currently our server never returns information about LivePatch plugin. As a workaround - add the information in getAvailablePlugins. This way you can add the plugin. In case you try publishing LivePatch (`$ appbuilder appmanager livesync android` for example), the server will check if you are able to use this feature and will fail in case you are not.

Remove the HACK when server is working correctly.
http://teampulse.telerik.com/view#item/292698


teobugslayer:
:+1: after reasonably green build

rosen-vladimirov:
just to add
sad_panda

PR #1082, opened 2015-06-01
Enable AndroidHardwareAcceleration when migrating to Cordova version greater than 4.0.0

Fatme:
Fixes http://teampulse.telerik.com/view#item/292884


Fatme:
Should be merged after this https://github.com/Icenium/Ice/pull/3774

teobugslayer:
:+1:

PR #1084, opened 2015-06-01
Fix prop print for properties which are not arrays

rosen-vladimirov:
When property is not array and configuration specific flag is used for prop print command, the value of the property is not shown in the output.
Also if the property does not have value in configuration specific data and user wants to see only this configuration, we should not show empty table for this property.
Fixes http://teampulse.telerik.com/view#item/292902


teobugslayer:
:+1:

PR #1085, opened 2015-06-01
Add authentication command

Fatme:
Add-signin and add-signup commands were removed in new version of sb and authentication command was introduced.
Fixes http://teampulse.telerik.com/view#item/292900


teobugslayer:
I think you need to also delete **add-signup.md**

ikoevska:
Would you, please, update the TeamPulse item for the SB update as well? Right now it says that only add-about is a new command.

teobugslayer:
:+1:

Fatme:
I've updated the teampulse item.

PR #1086, opened 2015-06-02
Fix plugin commands for renamed marketplace plugins

rosen-vladimirov:
When marketplace plugin is renamed we have some issues with our implementation:
* when plugin had been installed, we check it by name. It has been chaned, so if you try adding it again, we add it once more to your configuration files.
* when you try to remove the plugin by name, we cannot find it and we do not remove it from the .abproject files.

All these operations should be based on plugin identifiers, not plugin name. Add code to check if the passed name is an old name of a plugin. As in our *.abproject files we keep the ids of the plugins, not their names, when user passes old name, we can get the old plugin id and see if it matches any of the installed plugin ids.

Add version printing when marketplace plugin is added (for example: "Plugin streethawk was successfully added for debug configuration with version 1.5.5).

Fixes http://teampulse.telerik.com/view#item/292712
Fixes http://teampulse.telerik.com/view#item/292710


teobugslayer:
:+1: but notify QAs to fix their test due to changed output messages.

PR #1087, opened 2015-06-02
Show more descriptive error for obsolete {N} projects

rosen-vladimirov:
When project is using obsolete NativeScript version and user tries to execute **$ appbuilder mobileframework** inside project dir, an error message is shown: "Cannot find version <v> in the supported versions." Change this to explanation that the current version is obsolete and the project cannot be migrated, but we can still build it.

Fixes http://teampulse.telerik.com/view#item/292936

### DIFF
--- a/lib/services/nativescript-migration-service.ts
+++ b/lib/services/nativescript-migration-service.ts
@@ -74,10 +74,11 @@ export class NativeScriptMigrationService implements IFrameworkMigrationService 
 				return framework.displayName;
 			}
 
+			this.checkIsVersionObsolete(version).wait();
 			this.$errors.failWithoutHelp("Cannot find version %s in the supported versions.", version);
 		}).future<string>()();
 	}
-	
+
 	public onFrameworkVersionChanging(newVersion: string): IFuture<void> {
 		return (() => {
 			let projectDir = this.$project.getProjectDir().wait();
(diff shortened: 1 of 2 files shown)


lib/services/nativescript-migration-service.ts
```diff
@@ -74,6 +74,10 @@ export class NativeScriptMigrationService implements IFrameworkMigrationService
 				return framework.displayName;
 			}
 
+			if(_.any(this.getObsoleteVersions().wait(), obsoleteVersion => obsoleteVersion === version)) {
+				this.$errors.failWithoutHelp(`Your project is using obsolete framework version - ${version}. You can still build your project, but you cannot migrate to a newer version.`);
```
teobugslayer: Maybe add a message how the user can remedy the situation. Something like "To use newer version of NativeScript, re-create the project from scratch. See the docs (if we have any)"

rosen-vladimirov: ping @ikoevska for this message

ikoevska: Your project targets NativeScript ${version}. This version is obsolete and cannot be migrated from the command line. If you want to migrate your project, create a new project and copy over your code and resources.

teobugslayer:
:+1:

PR #1088, opened 2015-06-02
Allow parallel executions of many CLIs

rosen-vladimirov:
Currently when you try to use CLI in more than one terminal simultaneously, only the last process keeps working, while the other ones stop immediately when the next process starts. This is due to our code, which recreates the "cli" watch file. Add check if the file exists and do not recreate it in such case.

Fixes http://teampulse.telerik.com/view#item/292941


teobugslayer:
:+1:

PR #1089, opened 2015-06-02
Allow LivePatch plugin only in release configuration

rosen-vladimirov:
When user executes **$ appbuilder plugin add com.telerik.LivePatch** we automatically add it to both debug and release configurations. This plugin cannot work with LiveSync plugin enabled in debug mode, so we should enable LivePatch only for release.

Fixes http://teampulse.telerik.com/view#item/291396

### DIFF
--- a/lib/services/plugins-service.ts
+++ b/lib/services/plugins-service.ts
@@ -20,6 +20,10 @@ export class PluginsService implements IPluginsService {
 
 		return this._identifierToPlugin;
 	}
+	
+	private pluginsForbiddenConfigurations: IStringDictionary = {
+		"com.telerik.LivePatch": this.$projectConstants.DEBUG_CONFIGURATION_NAME
+	}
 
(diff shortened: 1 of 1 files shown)


lib/services/plugins-service.ts
```diff
@@ -20,6 +20,12 @@ export class PluginsService implements IPluginsService {
 
 		return this._identifierToPlugin;
 	}
+	
+	private get oneConfigurationOnlyPlugins(): IStringDictionary {
```
teobugslayer: Why is this a property and not a simple object?

rosen-vladimirov: Because the idea was to easily add other plugins which we want to have in only one configuration. But in the moment there aren't such plugins, so I can hardcode the id of the LivePatch plugin directly in addPlugin method.
What do you think about this?

teobugslayer:
:+1: after green build. Consider adding some unit tests as well.

PR #1092, opened 2015-06-03
Get all supported sb commands from .schema.json instead calling screenBuilder's listGenerators endpoint

Fatme:
Fixes http://teampulse.telerik.com/view#item/292965


rosen-vladimirov:
:+1:

tytodorov:
pr_ceco

PR #1093, opened 2015-06-03
Add LivePatch plugin to release only

rosen-vladimirov:
Currently LivePatch plugin is added to debug config instead of release. This is due to mistake in the code, which is setting configurations variable to be the forbiddenConfig. Fix this code.


endarova:
:+1:

Fatme:
:+1:

PR #1099, opened 2015-06-04
Migrate App_Resources directory in {N} projects

rosen-vladimirov:
App_Resources directory of NativeScript projects should be part of app directory. In 0.10.0, this directory was on the same level as app directory. Make sure to move it under app directory during project migration.
When checking all platform assets, make sure to update correct App_Resources directory. If the project is not 0.9.0 and ther's no App_Resources directory inside app directory, check for existing of App_Resources in the upper level and modify the App_Resoureces there.

Part of http://teampulse.telerik.com/view#item/293294

### DIFF
--- a/lib/project/nativescript-project.ts
+++ b/lib/project/nativescript-project.ts
@@ -5,6 +5,7 @@ import path = require("path");
 import util = require("util");
 import Future = require("fibers/future");
 import frameworkProjectBaseLib = require("./framework-project-base");
+import semver = require("semver");
 
 export class NativeScriptProject extends frameworkProjectBaseLib.FrameworkProjectBase implements Project.IFrameworkProject {
 	constructor(private $config: IConfiguration,
@@ -106,9 +107,16 @@ export class NativeScriptProject extends frameworkProjectBaseLib.FrameworkProjec
 		return (() => {
 			let appResourcesDir = this.$resources.appResourcesDir;
 			let appResourceFiles = this.$fs.enumerateFilesInDirectorySync(appResourcesDir);
+			// In 0.10.0 original template, App_Resources directory is not included in app directory.
+			let appResourcesHolderDirectory = path.join(projectDir, this.$projectConstants.NATIVESCRIPT_APP_DIR_NAME);
+			if(semver.eq(frameworkVersion, "0.9.0")  
+				|| (!this.$fs.exists(path.join(appResourcesHolderDirectory, this.$projectConstants.NATIVESCRIPT_APP_RESOURCES_DIR_NAME)).wait() 
+				&& this.$fs.exists(path.join(projectDir, this.$projectConstants.NATIVESCRIPT_APP_RESOURCES_DIR_NAME)).wait())) {
+				appResourcesHolderDirectory = projectDir;
+			}
 			appResourceFiles.forEach((appResourceFile) => {
 				let relativePath = path.relative(appResourcesDir, appResourceFile);
-				let targetFilePath = path.join(projectDir, relativePath);
+				let targetFilePath = path.join(appResourcesHolderDirectory, relativePath);
 				this.$logger.trace("Checking app resources: %s must match %s", appResourceFile, targetFilePath);
 				if (!this.$fs.exists(targetFilePath).wait()) {
 					this.printAssetUpdateMessage();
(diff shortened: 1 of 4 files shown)


lib/project/nativescript-project.ts
```diff
@@ -106,9 +106,15 @@ export class NativeScriptProject extends frameworkProjectBaseLib.FrameworkProjec
 		return (() => {
 			let appResourcesDir = this.$resources.appResourcesDir;
 			let appResourceFiles = this.$fs.enumerateFilesInDirectorySync(appResourcesDir);
+			// In 0.10.0 original template, App_Resources directory is not included in app directory.
+			let appResourcesHolderDirectory = path.join(projectDir, "app");
```
teobugslayer: Consider introducing a constant for "app"

lib/services/nativescript-migration-service.ts
```diff
@@ -97,12 +105,21 @@ export class NativeScriptMigrationService implements IFrameworkMigrationService
 				let projectType = this.$project.isTypeScriptProject().wait() ? NativeScriptMigrationService.TYPESCRIPT_ABBREVIATION : NativeScriptMigrationService.JAVASCRIPT_ABBREVIATION;
 				let pathToNewTnsModules = path.join(this.tnsModulesDirectoryPath, projectType, this.getFileNameByVersion(newVersion));
 				this.$fs.unzip(pathToNewTnsModules, tnsModulesProjectPath).wait();
+				if(!this.$fs.exists(appResourcesRequiredPath).wait() && this.$fs.exists(appResourcesObsoletePath).wait()) {
+					this.$logger.info(`Moving ${appResourcesObsoletePath} to ${appResourcesRequiredPath}`);
```
rosen-vladimirov: I'm not sure this should be info message. What do you think, @teobugslayer , @Fatme ?

teobugslayer:
:+1:

lib/services/nativescript-migration-service.ts
rosen-vladimirov: After discussion, we've decided to keep it as "info" message.

lib/project/nativescript-project.ts
```diff
@@ -106,9 +106,15 @@ export class NativeScriptProject extends frameworkProjectBaseLib.FrameworkProjec
 		return (() => {
 			let appResourcesDir = this.$resources.appResourcesDir;
 			let appResourceFiles = this.$fs.enumerateFilesInDirectorySync(appResourcesDir);
+			// In 0.10.0 original template, App_Resources directory is not included in app directory.
+			let appResourcesHolderDirectory = path.join(projectDir, this.$projectConstants.NATIVESCRIPT_APP_DIR_NAME);
+			if(frameworkVersion === "0.9.0" || (!this.$fs.exists(path.join(appResourcesHolderDirectory, this.$projectConstants.NATIVESCRIPT_APP_RESOURCES_DIR_NAME)).wait()
```
Mitko-Kerezov: you can use semver module here

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

lib/services/nativescript-migration-service.ts
```diff
@@ -97,12 +105,21 @@ export class NativeScriptMigrationService implements IFrameworkMigrationService
 				let projectType = this.$project.isTypeScriptProject().wait() ? NativeScriptMigrationService.TYPESCRIPT_ABBREVIATION : NativeScriptMigrationService.JAVASCRIPT_ABBREVIATION;
 				let pathToNewTnsModules = path.join(this.tnsModulesDirectoryPath, projectType, this.getFileNameByVersion(newVersion));
 				this.$fs.unzip(pathToNewTnsModules, tnsModulesProjectPath).wait();
+				if(!this.$fs.exists(appResourcesRequiredPath).wait() && this.$fs.exists(appResourcesObsoletePath).wait()) {
```
Fatme: We use this check more than once. Can we extract it to separate function?

teobugslayer:
run ci

Fatme:
:+1:  my suggestion above is not a merge stopper.

lib/services/nativescript-migration-service.ts
rosen-vladimirov: Currently it's quite difficult to share code between nativescirpt project and nativescript migration service and this will require a lot of refactoring, which I believe is not for this PR. So I'll keep it this way and fix it in the future.

tytodorov:
ok, then :+1:

PR #1103, opened 2015-06-05
Use correct App_Resources

rosen-vladimirov:
Currently we have App_Resources in our resources folder. It is created on dev-prepackage command by extracting App_Resources directory from Cordova.Blank template. Now we need App_Resources for NativeScript projects. With the new templates, the App_Resources folder is under app folder of {N} templates. In order to handle this, I've modified templatesService unpackAppResources to extract cordova and ns App_Resources.
Move App_Resources folder to be under Cordova and NativeScript directories. Also with the old code, we had App_Resources/App_Resources/Android..., now we have Cordova/App_Resources/Android...
Change the name of NATIVESCRIPT_APP_RESOURCES_DIR_NAME to APP_RESOURCES_DIR_NAME as this name is not specific for NativeScript projects.
Replace appResourcesDir property of resources with two new properties - cordovaAppResourcesDir and nativeScriptAppResourcesDir.
Inside prepackage, move the logic for unpacking app resources to be at the end as cordova migration data and nativescript migration data are deleting the directory where app resources are extracted.
Fixes http://teampulse.telerik.com/view#item/293506

### DIFF
--- a/lib/resource-loader.ts
+++ b/lib/resource-loader.ts
@@ -6,7 +6,8 @@ import helpers = require("./helpers");
 import util = require("util");
 
 export class ResourceLoader implements IResourceLoader {
-	constructor(private $fs: IFileSystem) { }
+	constructor(private $fs: IFileSystem, 
+		private $projectConstants: Project.IProjectConstants) { }
 
 	resolvePath(resourcePath: string): string {
 		return path.join(__dirname, "../resources", resourcePath);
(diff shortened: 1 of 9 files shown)


teobugslayer:
:+1:

lib/resource-loader.ts
```diff
@@ -20,8 +21,12 @@ export class ResourceLoader implements IResourceLoader {
 		return this.$fs.readJson(this.resolvePath(resourcePath));
 	}
 
-	public get appResourcesDir(): string {
-		return this.resolvePath("App_Resources");
+	public get cordovaAppResourcesDir(): string {
```
Fatme: Shouldn't this be in `cordova project`?

rosen-vladimirov: It's part of the resources, the directory itself is part of the resources and that's why I've placed it here. Placing it inside cordova project would mean it is part of any cordova project, while it is part of our own resources. 
That was my logic, but in case you insist, I'll try moving this to cordova project (not sure what kind of mess will happen).

Fatme: Or maybe we don't need separate functions for `cordova` and `nativescript` framework, but one function that get `frameworkName` as parameter and just ` return path.join(this.resolve(frameworkName), this.$projectConstants.APP_RESOURCES_DIR_NAME` . In this case it is ok to be here in `resource-loader` file.

Fatme:
:+1:

endarova:
:+1:

PR #1104, opened 2015-06-05
Filter plugins not supported for projects Cordova version

rosen-vladimirov:
In case plugin requires Cordova 3.5.0, but current project targets 3.2.0, adding such plugin will break the build. Prevent this by removing such plugins from getAvailablePlugins result. This will also hide the plugin from `appbuilder plugin --available` output.
Add unit tests for the result of getAvailablePlugins and for livePatch plugin.

Fixes http://teampulse.telerik.com/view#item/292103


teobugslayer:
run ci

teobugslayer:
:+1: but consider going through plugins-service.ts and add all missing semi-colons.

PR #1105, opened 2015-06-08
Fix emulate command

Fatme:
http://teampulse.telerik.com/view#item/293581


Mitko-Kerezov:
:+1:

tytodorov:
:+1: After green build

PR #1108, opened 2015-06-09
Fix test command in package.json

rosen-vladimirov:
Currently our unit tests cannot be started on Windows as the .bin/_mocha file is shell script and some of its symbols cannot be resolved. Replace the call with call to javascript file.


Fatme:
:+1:

teobugslayer:
:+1:

PR #1109, opened 2015-06-10
Expose answers option in order to automate ScreenBuilder integration

Fatme:
http://teampulse.telerik.com/view#item/294296


teobugslayer:
:+1: after a green build

PR #1112, opened 2015-06-12
Pass CordovaPluginVariables on the command line

rosen-vladimirov:
When you want to automate plugin add, there's no way to automate the process of adding CordovaPluginVariables and prompter is shown for each of them. In order to simplify this, add `var` dashed option, which should work in the following way:
* `$ appbuilder plugin add dropbox --var.APP_KEY 1 --var.APP_SECRET 2` - sets the plugin variables APP_KEY to 1 and APP_SECRET to 2 in all configurations
* `$ appbuilder plugin add dropbox --var.debug.APP_KEY 1 --var.debug.APP_SECRET 2 --var.release.APP_KEY 3 --var.release.APP_SECRET 4` - in .debug.abproject sets APP_KEY to 1 and APP_SECRET to 2 and in .release.abproject sets APP_KEY to 3 and APP_SECRET to 4
* `$ appbuilder plugin add dropbox --var.APP_KEY 1 --var.APP_SECRET 2  --var.debug.APP_KEY 3` - in .debug.abproject sets APP_KEY to 3 and APP_SECRET to 2 and in .release.abproject sets APP_KEY to 1 and APP_SECRET to 4 - when plugin variable is passed both with config and without config, the value for config is used for the respective configuration, the other value is used for all other values.
* `$ appbuilder plugin add dropbox --var.APP_KEY 1` - sets APP_KEY to 1 in all configurations, and prompts user for APP_SECRET for all configurations
* `$ appbuilder plugin add dropbox --var.APP_KEY 1 --var.debug.APP_SECRET 2` - sets APP_KEY to 1 in all configurations, APP_SECRET to 2 in debug config and prompts user for APP_SECRET value for release configuration

Also keys passed to --var will work no matter of the casing, for example --var.DeBuG.App_KeY is treated the same way as --var.debug.APP_KEY and will be used to set APP_KEY in debug configuration to passed value.

Added unit tests for the described scenarios.

http://teampulse.telerik.com/view#item/292387

### DIFF
--- a/docs/man_pages/lib-management/plugin-add.md
+++ b/docs/man_pages/lib-management/plugin-add.md
@@ -8,6 +8,8 @@ Add plugins | `$ appbuilder plugin add <Name or ID> [--debug] [--release]`
 Add a specific version of a plugin | `$ appbuilder plugin add <Name or ID>@<Version> [--debug] [--release]`
 Add latest version of a plugin | `$ appbuilder plugin add <Name or ID> --latest [--debug] [--release]`
 Add default version of a plugin | `$ appbuilder plugin add <Name or ID> --default [--debug] [--release]`
+Add plugin and specify plugin variables | `$ appbuilder plugin add <Name or ID> [--var.<variable_id> <variable value> [--var.<variable_id> <variable value>]*>]`
+Add plugin and specify plugin variables per configuration | `$ appbuilder plugin add <Name or ID> [--var.<configuration>.<variable_id> <variable value> [--var.<configuration>.<variable_id> <variable value>]*>]`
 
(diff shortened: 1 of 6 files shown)


teobugslayer:
I am not sure about the * in the help - do we use it anywhere else?

:+1: for the code.

rosen-vladimirov:
@teobugslayer, 
Yes, we are using `*` in our help, I've already discussed its usage with @ikoevska . You can check the usage here https://github.com/Icenium/icenium-cli/blob/master/docs/man_pages/publishing/appmanager-upload-android.md 
Anyway I'll ping @ikoevska to check the content :)

PR #1114, opened 2015-06-15
Track user tenant information

rosen-vladimirov:
Add tracking of tenant edition and license. Use special names of the edition, similar to Mist.
Implements http://teampulse.telerik.com/view#item/293312

### DIFF
--- a/lib/login.ts
+++ b/lib/login.ts
@@ -14,10 +14,20 @@ export class UserDataStore implements IUserDataStore {
 	private cookies: IStringDictionary;
 	private user: any;
 
+	private static APPBUILDER_EDITION_TYPE: IStringDictionary = {
+		"Starter": "AppBuilder Starter Edition",
+		"Developer": "AppBuilder Developer Edition",
+		"Business": "AppBuilder Business Edition",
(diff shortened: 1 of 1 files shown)


lib/login.ts
```diff
@@ -14,10 +14,20 @@ export class UserDataStore implements IUserDataStore {
 	private cookies: IStringDictionary;
 	private user: any;
 
+	private static APPBUILDER_EDITION_TYPE: IStringDictionary = {
```
ligaz: Can those taken from the server?

rosen-vladimirov: Unfortunately no :(
We receive only the enum from the Server (https://github.com/Icenium/Ice/blob/master/Common/Common.Data/Authentication/EditionType.cs) while we have to send the "Description" of each item. Same logic is implemented in Mist:
https://github.com/Icenium/Ice/pull/3854/files

ligaz: We can definitely change this on the server then.

teobugslayer:
:+1: But we may consider adding a new APi on the server to give all clients consistent names for these./

PR #1115, opened 2015-06-17
Parse --var option when variable has dots

rosen-vladimirov:
In case a plugin has variable with dot ("."), you cannot pass it on the command line with --var option as each dot is converted to new inner object by yargs. Bypass this behavior, by converting the yargs object to one that matches our required behavior.

Add unit tests.


rosen-vladimirov:
run ci

teobugslayer:
:+1: and the tests are super-awesome!

PR #1117, opened 2015-06-17
Remove client option and fix analytics tracking

rosen-vladimirov:
Client option was used in analytics tracking only, but it has been replaced there, so we do not need it anymore.


Fatme:
:+1:

PR #1120, opened 2015-06-18
Introducing --var

ikoevska:
When reviewing this, keep in mind the following:

* There is a known issue that Rosko has committed to fix. In some cases, the CLI shows `[*` instead of the expected `*`.
* I have deliberately dropped or avoided explanations of the various `--configuration` and `--var.configuration` combos since this is not a scenario that we want to advertise. Users should use either `--configuration --var` or just `--var.configuration` to set something in a particular configuration.
* Currently, if you have enabled a plugin in one configuration but run `plugin configure --var.the-other-configuration` or `plugin configure --var` we enable the plugin for the other configuration as well. This is an erroneous behavior and will be fixed.

### DIFF
--- a/docs/man_pages/lib-management/plugin-add.md
+++ b/docs/man_pages/lib-management/plugin-add.md
@@ -1,17 +1,22 @@
 plugin add
 ==========
 
-Usage | Synopsis
-------|-------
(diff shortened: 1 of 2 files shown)


teobugslayer:
:+1:

rosen-vladimirov:
run ci

PR #1121, opened 2015-06-19
Fix markdown links in the console

rosen-vladimirov:
Fix the regex for markdown links, so `[\*\*](http://....)` can be shown correctly as \*\*


rosen-vladimirov:
run ci

teobugslayer:
:+1: after green build

PR #1122, opened 2015-06-22
Add dashedOptions for each command

rosen-vladimirov:
Each command can use specific dashed options (`-- avd <name>` for example is valid for one command, but not for other). Add dashedOptions parameter to ICommand interface and make sure to use it for validation when it exists. In case there's no dashedOptions paramter, use the common options.

Use correct types for yargs - "string", "boolean", etc. instead of 0, 1, 2 which were the result of our enum. Replace the enum with class and remove code, which we do not need anymore as yargs is doing its work as expected.
After yargs is updated to latest version, some fixes have to be applied - now when `yargs.options(opts).argv` is passed, it returns all members of `opts`, but the ones that are not part of process.argv have value undefined.
Fix unit tests based on the changes and remove some, which were testing our hacks. (+1 squashed commits)

### DIFF
--- a/test/commands-service.ts
+++ b/test/commands-service.ts
@@ -9,6 +9,9 @@ let configFile = require("../lib/config");
 import util = require("util");
 let assert = require("chai").assert;
 let commandParams = require("../lib/common/command-params");
+import commonOptionsLib = require("../lib/common/options");
+let optionsLib = require("../lib/options");
+import hostInfoLib = require("../lib/common/host-info");
 
 let isCommandExecuted: boolean;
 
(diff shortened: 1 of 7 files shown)


test/commands-service.ts
```diff
@@ -358,8 +379,50 @@ describe("commands service", () => {
...
+	describe("tryExecuteCommand when command has dashed options", () => {
+		let commandsService: any;
+		let testInjector: IInjector;
+		beforeEach(() => {
+			testInjector = createTestInjector();
+			isCommandExecuted = false;
+		});
+
+		it("does not execute command when it has its own dashed options and invalid one is passed", () => {
+			isCommandExecuted = false;
```
teobugslayer: I think this line is not needed.

test/commands-service.ts
```diff
@@ -358,8 +379,50 @@ describe("commands service", () => {
...
+			process.argv.push("--availableDevices");
+			testInjector.register("options", optionsLib.Options);
+			commandsService = testInjector.resolve("commandsService");
+			commandsService.tryExecuteCommand("commandWithDashedOptions", []).wait();
+			assert.isFalse(isCommandExecuted);
+			process.argv.pop();
+		});
+
+		it("executes command when it has its own dashed options and one of them is passed", () => {
+			isCommandExecuted = false;
```
teobugslayer: Again not needed

test/commands-service.ts
rosen-vladimirov: yes, I'll remove it from all tests, thanks for noticing this :)

teobugslayer:
run ci

teobugslayer:
:+1:

PR #1126, opened 2015-06-29
Fix creation of SB when simulator start fails

rosen-vladimirov:
When you create SB project and for some reason starting the simulator fails, the project is deleted. Make sure the start the simulator outside of the try-catch, so even if it fails, the project will not be deleted when the simulator fails.
Fixes http://teampulse.telerik.com/view#item/296228


teobugslayer:
:+1:

PR #1128, opened 2015-06-29
--answers option for SB commands

ikoevska:
IMPORTANT: To check the contents at the links replace docs with testdocs. Here's the address of the article: http://testdocs.telerik.com/platform/appbuilder/creating-your-project/screen-builder-automation


teobugslayer:
:+1: but I suggest we consult with the SB team to verify the contents on docs.telerik.com. ping @lini

PR #1130, opened 2015-06-30
Use correct semver version in package.json

rosen-vladimirov:
npm 2.11.2 introduced and issue when the version in package json has "-". Such packages cannot be installed directly from npmjs.org.
Fix this by adding buildVersion property in our package.json and set there the build number from Jenkins. This way the version in package.json will remain correct. Add "setPackageName" grunt task, which renames the `appbuilder-<version>.tgz` to `appbuilder-<version>-<buildVersion>.tgz`

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -112,5 +112,9 @@ export class StaticConfig extends staticConfigBaseLib.StaticConfigBase implement
 	public get HTML_CLI_HELPERS_DIR(): string {
 		return path.join(__dirname, "../docs/helpers");
 	}
+	
+	public get pathToPackageJson(): string {
+		return path.join(__dirname, "..", "package.json");
+	}
 }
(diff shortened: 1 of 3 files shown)


lib/config.ts
```diff
@@ -112,5 +112,9 @@ export class StaticConfig extends staticConfigBaseLib.StaticConfigBase implement
 	public get HTML_CLI_HELPERS_DIR(): string {
 		return path.join(__dirname, "../docs/helpers");
 	}
+	
+	public get PATH_TO_PACKAGE_JSON(): string {
```
teobugslayer: Why is this marked as a constant? It is a property

teobugslayer:
:+1:

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #1134, opened 2015-07-02
Do not require latest version for OnPremise

rosen-vladimirov:
CLI has built-in mechanism for checking if the current version is the latest one. If it is not, CLI does not execute you any http requests (almost all commands have such requests). For OnPremise we do not want this check, so make sure to skip this behavior.

Fixes http://teampulse.telerik.com/view#item/296618


rosen-vladimirov:
run ci

Mitko-Kerezov:
:+1:

PR #1135, opened 2015-07-02
Respect default enabled plugins

mbektchiev:
fixes [Turn on Whitelist plugin when migrating to Cordova 4.0.0](http://teampulse.telerik.com/view#item/295506)


lib/services/cordova-migration-service.ts
```diff
@@ -17,7 +17,7 @@ class MigrationData {
 	constructor(public renamedPlugins: RenamedPlugin[],
 		public supportedVersions: string[],
 		public integratedPlugins: { [version: string]: string[] },
-		public supportedFrameworkVersions: IFrameworkVersion[],
+		public supportedFrameworkVersions: IFrameworkVersion[],
```
rosen-vladimirov: additional spaces/tabs at the end of the row

rosen-vladimirov:
:+1:  great work

PR #1136, opened 2015-07-02
Fix swagger not generating enums properly

Mitko-Kerezov:
Includes fixes for two bugs
* if an enum is encountered an even amount of numbers on a given endpoint definition it is not added to the end .d.ts result
* if an enum is encountered in multiple endpoint definition it is added as a duplicate

### DIFF
--- a/lib/swagger/service-contract-generator.ts
+++ b/lib/swagger/service-contract-generator.ts
@@ -19,6 +19,7 @@ export class ServiceContractGenerator implements Server.IServiceContractGenerato
 
 	constructor(private $serviceContractProvider: Server.IServiceContractProvider) {
 		this.tsTypeSystemHelpers = new TSTypeSystemHelpersLib.TSTypeSystemHelpers();
+		this.pendingModels = {};
 	}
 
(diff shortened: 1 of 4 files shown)


Mitko-Kerezov:
run ci

lib/swagger/service-contract-generator.ts
```diff
@@ -161,13 +162,12 @@ export class ServiceContractGenerator implements Server.IServiceContractGenerato
 	}
 
 	private ensureEnumAdded(allowableValues: Swagger.IModelPropertyValue) {
-		let enumBlock: Swagger.IBlock;
 		let typeName = this.tsTypeSystemHelpers.translate(allowableValues.valueType);
 		if (!this.pendingModels[typeName]) {
-			enumBlock = new codeEntityLib.Block(util.format("const enum %s", typeName));
+			let enumBlock = new codeEntityLib.Block(util.format("const enum %s", typeName));
```
rosen-vladimirov: util.format?

rosen-vladimirov:
:+1:

PR #1138, opened 2015-07-03
Switching to kebab-case for two-worded options (where working)

ikoevska:
http://teampulse.telerik.com/view#item/295423


teobugslayer:
:+1: Great work! By the way, please file a bug report if you find an option which does not work in kebab form.

ikoevska:
I did, the only one for the AB CLI was --appid for the create commands =) (it might not be expected to work though). And for the NS CLI -> justlaunch

PR #1142, opened 2015-07-06
Fix analytics issues

rosen-vladimirov:
- Fix behavior that is trying to login the user when tracking command - in fact we should first check if the user is logged in and if not, do not try to login.
- Make sure commands are tracked, but for some of them we should not ask the user for confirmation of usage reporting (for example help command). So add new command property - `disableAnalyticsConsentCheck` and set it to true for help, usage-reporting and error-reporting commands. Make sure to prevent commandsService for checking the consent when this property is true.

### DIFF
--- a/.settings/launch.json
+++ b/.settings/launch.json
@@ -1,0 +1,41 @@
+{
+	"version": "0.1.0",
+	// List of configurations. Add new configurations or edit existing ones.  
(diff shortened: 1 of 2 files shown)


Fatme:
:+1:  after green/yellow build

PR #1143, opened 2015-07-07
Integrate ScreenBuilder v0.0.8

Mitko-Kerezov:
Latest changes include
* JSDO data source support
* Unique default values

Implements [#294209](http://teampulse.telerik.com/view#item/294209)


Fatme:
:+1:

PR #1144, opened 2015-07-08
Export solutions

rosen-vladimirov:
According to our discussions, we would like to be able to export full solutions, not only one project. So the behavior of `appbuilder cloud export` command should be:
* `$ appbuilder cloud export` - prompts to select solution from which to export project. After solution is selected, prompt the user to select project. In this prompt there's option - export solution, when it is used, the full solution is exported.
* `$ appbuilder cloud export Sln1` - Exports the whole solution. In case the solution does not have any projects - show error.
* `$ appbuilder cloud export 1` - Exports the selected solution (first project listed by `$ appbuilder cloud`). In case the solution does not have any projects - show error. 
* `$ appbuilder cloud export Sln1 Proj1` - Exports Proj1 of Sln1. If any of them does not exists, throw error.
* `$ appbuilder cloud export 1 Proj1` - Exports Proj1 from the first solution listed by `$ appbuilder cloud`. If any of them does not exists, throw error.
* `$ appbuilder cloud export Sln1 1` - Exports first project of Sln1 listed by `$ appbuilder cloud`. If any of them does not exists, throw error.
* `$ appbuilder cloud export 1 1` - Exports first project of the first solution listed by `$ appbuilder cloud`. If any of them does not exists, throw error.

Move the logic of exporting to `remoteProjectService`. Keep the logic what to export in the command. Add check if the command `appbuilder cloud export` is executed with parameters in non-interactive console. In this case it will fail, as we do not know what to export.

http://teampulse.telerik.com/view#item/295399

### DIFF
--- a/lib/services/remote-projects-service.ts
+++ b/lib/services/remote-projects-service.ts
@@ -9,7 +9,7 @@ import temp = require("temp");
 export class RemoteProjectService implements IRemoteProjectService {
 	private clientSolutions: Server.TapSolutionData[];
 	private clientProjectsPerSolution: IDictionary<Server.IWorkspaceItemData[]> = {};
-
+	
 	constructor(private $server: Server.IServer,
 				private $userDataStore: IUserDataStore,
 				private $serviceProxy: Server.IServiceProxy,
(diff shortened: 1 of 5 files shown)


rosen-vladimirov:
NOTE: Some additional tests must be added and I would like to simplify the tests for cloud-export command. This will be handled in separate PR.

lib/services/remote-projects-service.ts
```diff
@@ -94,5 +134,39 @@ export class RemoteProjectService implements IRemoteProjectService {
...
+		return (() => {
+			try {
+				// if there is no .abproject when exporting, we must be dealing with a cordova project, otherwise everything is set server-side
+				let projectFile = path.join(projectDir, this.$projectConstants.PROJECT_FILE);
+				if(!this.$fs.exists(projectFile).wait()) {
+					let properties = this.getProjectProperties(remoteSolutionName, remoteProjectName).wait();
+					this.$project.createProjectFile(projectDir, properties).wait();
+				}
+			} catch(e) {
+				this.$logger.warn("Couldn't create project file: %s", e.message);
```
Mitko-Kerezov: This can be a TS template string

Mitko-Kerezov:
:+1:

Fatme:
:+1:

PR #1146, opened 2015-07-08
Fix mobileframework-set references in help

Mitko-Kerezov:


docs/man_pages/project/configuration/edit-configuration.md
```diff
@@ -28,7 +28,7 @@ WARNING: This command is not applicable to mobile website projects. To view the
 Command | Description
 ----------|----------
 [mobileframework](mobileframework.html) | Lists all supported versions of the current development framework.
-[mobileframework](mobileframework.html) | Sets the selected development framework version for the project.
+[mobileframework&nbsp;set](mobileframework-set.html) | Sets the selected development framework version for the project.
```
rosen-vladimirov: is there a particular reason to use `&nbsp`, `prop add`, `prop print`, etc. are not using it

ikoevska:
:+1:  Still, I wonder what happened here so that it turned out this way o.O

docs/man_pages/project/configuration/edit-configuration.md
Mitko-Kerezov: Now that you mention it I don't think it's necessary

PR #1148, opened 2015-07-08
Support {N} 1.1.2

rosen-vladimirov:
Add support fore {N} 1.1.2. During migration from older version to 1.1.2, we have to create package.json file (take it from nativescript resources directory). Make sure to keep this file when downgrading from 1.1.2 to older version.

http://teampulse.telerik.com/view#item/294271

> NOTE: This PR will fail until BpcTooling and Ice are not merged

### DIFF
--- a/lib/services/nativescript-migration-service.ts
+++ b/lib/services/nativescript-migration-service.ts
@@ -4,16 +4,19 @@
 import Future = require("fibers/future");
 import path = require("path");
 import helpers = require("./../helpers");
+import semver = require("semver");
 
 export class NativeScriptMigrationService implements IFrameworkMigrationService {
 	private static TYPESCRIPT_ABBREVIATION = "TS";
 	private static JAVASCRIPT_ABBREVIATION = "JS";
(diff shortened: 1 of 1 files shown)


lib/services/nativescript-migration-service.ts
```diff
@@ -4,16 +4,19 @@
...
 	private static TYPESCRIPT_ABBREVIATION = "TS";
 	private static JAVASCRIPT_ABBREVIATION = "JS";
 	private static SUPPORTED_LANGUAGES = [NativeScriptMigrationService.JAVASCRIPT_ABBREVIATION, NativeScriptMigrationService.TYPESCRIPT_ABBREVIATION];
 	private static REMOTE_NATIVESCRIPT_MIGRATION_DATA_FILENAME = "NativeScript.json";
+	private static PACKAGE_JSON_FILE_NAME = "package.json";
 	private nativeScriptResourcesDir: string;
 	private nativeScriptMigrationFile: string;
 	private tnsModulesDirectoryPath: string;
 	private remoteNativeScriptResourcesPath: string;
+	private nativeScriptDefaultPackageJsonFile: string;
```
Fatme: Do we need default?

Fatme: Also we're inside nativescript-migration-service, do we need `nativescript` prefix?

Fatme:
:+1:

PR #1150, opened 2015-07-09
Update NativeScript FrameworkVersion in tests

Mitko-Kerezov:
Change the FrameworkVersion property so that it matches that on the server tempalate


Fatme:
:+1:

PR #1153, opened 2015-07-13
Enhance doctor command to report about missing Git

teobugslayer:
Implements http://teampulse.telerik.com/view#item/296931


rosen-vladimirov:
:+1:

PR #1155, opened 2015-07-13
Fix Invalid property name 'undefined' message

Mitko-Kerezov:
When running `$ appbuilder prop add` or any other prop-related command without a property name the message Invalid property name 'undefined' is displayed.
Fixes [#297059](http://teampulse.telerik.com/view#item/297059)


rosen-vladimirov:
:+1:

rosen-vladimirov:
but is this for master?
maybe it should target release branch

PR #1156, opened 2015-07-13
Fix Invalid property name 'undefined' message

Mitko-Kerezov:
When running `$ appbuilder prop add` or any other prop-related command without a property name the message Invalid property name 'undefined' is displayed.
Fixes http://teampulse.telerik.com/view#item/297059
Retarget from https://github.com/Icenium/icenium-cli/pull/1155


rosen-vladimirov:
:+1:

PR #1154, opened 2015-07-13
Fix screenbuilder help `Related Commands`

Mitko-Kerezov:
Ping @ikoevska

PR #1157, opened 2015-07-13
Fix screenbuilder help `Related Commands`

Mitko-Kerezov:
Ping @ikoevska
This is a retarget from https://github.com/Icenium/icenium-cli/pull/1154


ikoevska:
Looking good :+1:

PR #1159, opened 2015-07-15
Print all error messages upon screenbuilder failure

Mitko-Kerezov:
Upon a failure in any screenbuilder command an error object is thrown which contains multiple levels of nested errors. We should print all of the errors' messages instead of just root-level ones.
Fixes [#296877](http://teampulse.telerik.com/view#item/296877)


Mitko-Kerezov:
Ping @Fatme @rosen-vladimirov for a quick review

rosen-vladimirov:
:+1:

PR #1161, opened 2015-07-16
Move screebuilder cache under sb directory

rosen-vladimirov:
On Mac and Linux, move SB cache under sb directory. This will fix the issue that during first downloading of ScreenBuilder, CLI is deleting its own cache and after that a browser is opened, waiting for login information.

Fixes http://teampulse.telerik.com/view#item/297519


Mitko-Kerezov:
:+1:

PR #1162, opened 2015-07-17
Add editorconfig file

Mitko-Kerezov:
[Editorconfig](http://editorconfig.org/) helps a bit with spacing like for example removing trailing whitespaces, inserting a newline at the end of the file and so on
Editors which currently support editorconfig can be found [here](http://editorconfig.org/#download)


rosen-vladimirov:
:+1:

PR #1163, opened 2015-07-17
Release the terminal when starting android emulator

rosen-vladimirov:
Release the terminal when starting android emulator. Fixes: http://teampulse.telerik.com/view#item/297523


Mitko-Kerezov:
This code should be refactored so that NativeScript-specific logic stays within its own CLI (e. g. justlaunch option is {N} - specific).
Due to pressing release dates and some mixups in lib/common this cannot be done now but is a must for the future
:+1:

PR #1165, opened 2015-07-17
Fix editing of config files for {N} projects

rosen-vladimirov:
Fix editing of AndroidManifest and Info.plist for {N} projects. Currently we were creating them on the root level, but they should be inside app/App_Resources dir.

Fixes http://teampulse.telerik.com/view#item/297518


Mitko-Kerezov:
:+1:

PR #1166, opened 2015-07-17
Fix error message of build commands

rosen-vladimirov:
When build command is executed outside of project dir, its execute logic is triggered and the hook that's attached to it is throwing an error. This error is caught by our uncaughtException listener, which prints message with call stack and exits the process. We have catch for this code, but it loosk like the problem is with the two fibers working in parallel.

**Also update common  lib, where the following change is applied:**
In case command does not have default command and arguments are not passed, we are trying to execute default command, which is not expected. Make sure not to return true in this method, when there's no default command.

Fixes: http://teampulse.telerik.com/view#item/297504

### DIFF
--- a/lib/commands/build.ts
+++ b/lib/commands/build.ts
@@ -1,11 +1,35 @@
 ///<reference path="../.d.ts"/>
 "use strict";
 
-export class BuildAndroidCommand implements ICommand {
-	constructor(private $buildService: Project.IBuildService,
-		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants) { }
+import Future = require("fibers/future");
+import assert = require("assert");
 
+class BuildCommandBase implements ICommand {
+	constructor(private $project: Project.IProject,
+		private $errors: IErrors) { }
+	
 	allowedParameters: ICommandParameter[] = [];
+	
+	execute(args: string[]): IFuture<void> {
+		assert.fail("","", "You should never get here. Please contact Telerik support and send the output of your command, executed with `--log trace`.");
+		return Future.fromResult();
+	}
+	
+	canExecute(args: string[]): IFuture<boolean> {
+		return (() => {
+			this.$project.ensureProject();
+			if(args.length) {
+				this.$errors.fail("This command doesn't accept parameters.");
+			}
+			return true;
+		}).future<boolean>()();
+	}
+}
+export class BuildAndroidCommand extends BuildCommandBase {
+	constructor(private $buildService: Project.IBuildService,
+		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
+		$project: Project.IProject,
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,17 @@
+# Manage common editor configurations
+# set for example indentation tab and space
+# VS Extension: https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328
+# more extensions and plugins for your favorite editor here: http://editorconfig.org/#download
+
+; Top-most EditorConfig file
+root = true
+
+; 4-column space indentation
```
rosen-vladimirov: remove this file

lib/commands/build.ts
```diff
@@ -1,11 +1,34 @@
...
+import Future = require("fibers/future");
 
+class BuildCommandBase implements ICommand {
+	constructor(private $project: Project.IProject,
+		private $errors: IErrors) { }
+	
 	allowedParameters: ICommandParameter[] = [];
+	
+	execute(args: string[]): IFuture<void> {
+		this.$errors.failWithoutHelp("You should never get here. Please contact Telerik support and send the output of your command, executed with `--log trace`.");
```
teobugslayer: This should be an assert.

teobugslayer:
:+1:

lib/commands/build.ts
```diff
@@ -1,11 +1,35 @@
...
+import assert = require("assert");
 
+class BuildCommandBase implements ICommand {
+	constructor(private $project: Project.IProject,
+		private $errors: IErrors) { }
+	
 	allowedParameters: ICommandParameter[] = [];
+	
+	execute(args: string[]): IFuture<void> {
+		assert.fail("","", "You should never get here. Please contact Telerik support and send the output of your command, executed with `--log trace`.");
```
Fatme: Do we really need this assert?

rosen-vladimirov: yes, in case of future refactoring or new build command that extends BuildCommandBase, this assert will show the error.
We are using such asserts/errors.fail in other places as well :)

teobugslayer: I also think that the assert is a good idea, as the user should not be able to put the program into a state which executes this code, but we as developers can.

ligaz:
I would like to suggest a refactoring that will eliminate the need for those `Build<PlatformName>Command`s. We can have a generic one and just register three of those in the injector. I think we have done this in the past a couple of times.

rosen-vladimirov:
Yes, we've done this for hybrid, native and website subcommands. I agree with your idea for the same approach about android, ios and wp8 subcommands  (in fact I've started the refactoring), but I've decided to leave it for master branch. I'll log a separate dev story for this refactoring :)

PR #1167, opened 2015-07-20
Disable WP8 related commands for OnPremise

rosen-vladimirov:
WP8 related commands should be disabled for On Premise. Disable the following commands:
* `appmanager upload wp8`
* `build wp8`
* `deploy wp8`
* `emulate wp8`
* `livesync wp8`

Change `appbuilder appmanger livesync` command to exclude WP8 when listing available platforms. Also check if the user had specified WP8 platform, for example:
`appbuilder appmanger livesync android wp8` and fail if we are in On Premise scenario.

Do not allow editing of wp8-manifest and wp8-config files in OnPremise scenario.

Fixes http://teampulse.telerik.com/view#item/297596

### DIFF
--- a/lib/commands/appmanager-livesync.ts
+++ b/lib/commands/appmanager-livesync.ts
@@ -8,21 +8,27 @@ export class AppManagerLiveSyncCommand implements ICommand {
 		private $mobileHelper: Mobile.IMobileHelper,
 		private $appManagerService: IAppManagerService,
 		private $errors: IErrors,
-		private $logger:ILogger) { }
+		private $logger:ILogger,
+		private $config: Config.IConfig) { }
 
 	public execute(args: string[]): IFuture<void> {
 		return ((): void => {
+			let windowsPhonePlatformName = this.$mobileHelper.normalizePlatformName("WP8");
 			if(!args || args.length === 0) {
-				let options = this.$mobileHelper.platformNames.concat(AppManagerLiveSyncCommand.ALL_PLATFORMS_OPTION);
-				let selectedPlatform = this.$prompter.promptForChoice("This command will publish a new update version to AppManager. Please select platform?", options).wait();
+				let availablePlatforms = this.$config.ON_PREM ? _.without(this.$mobileHelper.platformNames, windowsPhonePlatformName) : this.$mobileHelper.platformNames;
+				let selectionOptions = availablePlatforms.concat(AppManagerLiveSyncCommand.ALL_PLATFORMS_OPTION);
+				let selectedPlatform = this.$prompter.promptForChoice("This command will publish a new update version to AppManager. Please select platform?", selectionOptions).wait();
 				if(selectedPlatform === AppManagerLiveSyncCommand.ALL_PLATFORMS_OPTION) {
-					this.$appManagerService.publishLivePatch(this.$mobileHelper.platformNames).wait();
+					this.$appManagerService.publishLivePatch(availablePlatforms).wait();
 				} else {
 					this.$appManagerService.publishLivePatch([selectedPlatform]).wait();
 				}
 			} else {
 				// make sure each platform is specified only once
 				let platforms = _.keys(_.groupBy(args, arg => this.$mobileHelper.normalizePlatformName(arg)));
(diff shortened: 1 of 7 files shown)


lib/commands/appmanager-livesync.ts
```diff
@@ -8,21 +8,27 @@ export class AppManagerLiveSyncCommand implements ICommand {
...
-					this.$appManagerService.publishLivePatch(this.$mobileHelper.platformNames).wait();
+					this.$appManagerService.publishLivePatch(availablePlatforms).wait();
 				} else {
 					this.$appManagerService.publishLivePatch([selectedPlatform]).wait();
 				}
 			} else {
 				// make sure each platform is specified only once
 				let platforms = _.keys(_.groupBy(args, arg => this.$mobileHelper.normalizePlatformName(arg)));
+				if(this.$config.ON_PREM && _.contains(platforms, windowsPhonePlatformName)) {
+					this.$errors.failWithoutHelp(`You cannot upload patch for '${windowsPhonePlatformName}' platform.`);
```
rosen-vladimirov: ping @ikoevska  for the error message

ikoevska:
```
You cannot upload updates for '${windowsPhonePlatformName}'.
```

lib/commands/appmanager-livesync.ts
```diff
@@ -8,21 +8,27 @@ export class AppManagerLiveSyncCommand implements ICommand {
 		private $mobileHelper: Mobile.IMobileHelper,
 		private $appManagerService: IAppManagerService,
 		private $errors: IErrors,
-		private $logger:ILogger) { }
+		private $logger:ILogger,
+		private $config: Config.IConfig) { }
 
 	public execute(args: string[]): IFuture<void> {
 		return ((): void => {
+			let windowsPhonePlatformName = this.$mobileHelper.normalizePlatformName("WP8");
```
Fatme: Do you need to call `normalizePlatformName` here?

rosen-vladimirov: My idea was to make sure that even if we change the value of  WP8 in mobileHelper.platforms, this code will still work. In case I put "WP8" here and later we decide to change the value in mobileHelper.platforms, we'll fail here.

lib/commands/appmanager-livesync.ts
```diff
@@ -8,21 +8,27 @@ export class AppManagerLiveSyncCommand implements ICommand {
...
-					this.$appManagerService.publishLivePatch(this.$mobileHelper.platformNames).wait();
+					this.$appManagerService.publishLivePatch(availablePlatforms).wait();
 				} else {
 					this.$appManagerService.publishLivePatch([selectedPlatform]).wait();
 				}
 			} else {
 				// make sure each platform is specified only once
 				let platforms = _.keys(_.groupBy(args, arg => this.$mobileHelper.normalizePlatformName(arg)));
+				if(this.$config.ON_PREM && _.contains(platforms, windowsPhonePlatformName)) {
+					this.$errors.failWithoutHelp(`You cannot upload updates for '${windowsPhonePlatformName}'.`);
```
teobugslayer: I think that the full platform name is better in the message. Contrast:
`You cannot upload updates for WP8` and `You cannot upload updates for Windows Phone`.

teobugslayer:
run ci

lib/commands/appmanager-livesync.ts
ikoevska: The full platform name is better indeed. Doesn't '${windowsPhonePlatformName}' stand for it o.O?

rosen-vladimirov: Nope, it stands for WP8...
So, should I change it?

ikoevska: Yes.

teobugslayer:
:+1: after changing the message.

lib/commands/appmanager-livesync.ts
Fatme: Yes, I know this, my idea was to use `devicePlatformsConstants`.

PR #1168, opened 2015-07-20
Parse command line values as strings

rosen-vladimirov:
Yargs will parse all non-hyphenated values as strings in case we tell we want this. So, let's declare we want this and remove our hack, that's doing it. This will also fix the case when numbers like 8.0 are passed on the console. Yargs is giving them to us as "8", but we need the full value. With this fix, we'll receive "8.0".

Remove the hack for WPSdk property as it is not required anymore.

The fix is required for http://teampulse.telerik.com/view#item/296803

### DIFF
--- a/lib/services/project-properties-service.ts
+++ b/lib/services/project-properties-service.ts
@@ -106,13 +106,6 @@ export class ProjectPropertiesService implements IProjectPropertiesService {
 				this.$errors.fail("Unknown property update mode '%s'", mode);
 			}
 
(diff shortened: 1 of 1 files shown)


teobugslayer:
run ci

teobugslayer:
:+1: but let's wait for a green build

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #1169, opened 2015-07-20
Update lib common to support bom encodings

ispasov:


ispasov:
:+1: from @rosen-vladimirov

PR #1170, opened 2015-07-20
Read package.json encoded with BOM

rosen-vladimirov:
When package.json is changed from On premise scripts, it is encoded with BOM, so requiring it is failing with error. Use our readJson implementation which is able to read such file without errors.


rosen-vladimirov:
:+1:  in common lib - only SHA is updated here

PR #1171, opened 2015-07-21
Prevent deploy on iOS device with lower iOSDeploymentTarget

rosen-vladimirov:
When iOSDeploymentTarget is set to 8.4 for example, but user is trying to deploy on iOS device with 8.1.2, we must fail as different issues occur (on Windows we say "Successfully deployed", but nothing is deployed, while on Mac the deployment freezes). Fix this behavior by showing error when device's OS is with lower version, but you can still deploy on other iOS devices.

Fixes http://teampulse.telerik.com/view#item/297476


Fatme:
:+1:

PR #1172, opened 2015-07-21
Fix adb spawn when path contains spaces

rosen-vladimirov:
In case user does not have Android SDK installed, we are trying to execute adb from CLI resources. This fails in case `$ <CLI> device` command or `$ <CLI> deploy <platform>` is executed as the "exec" cannot resolve the path with spaces.

Also fix the check dependencies of AndroidEmulatorService in order to fail with correct error message when something is not configured properly (currently it is failing with ENOENT error).

Fixes http://teampulse.telerik.com/view#item/297604


teobugslayer:
:+1:

teobugslayer:
run ci

PR #1174, opened 2015-07-23
When evaluating glob patterns with wild cards do not skip .files

teobugslayer:
For example, test patern test/\*\*/\* must exclude test/.file

Fixes http://teampulse.telerik.com/view#item/297858


rosen-vladimirov:
:+1:

ndinev:
:+1:

PR #1175, opened 2015-07-24
Fix deploy on iOS devices

rosen-vladimirov:
We should not validate iOSDeploymentTarget when this property is not set in .abproject (it is not mandatory).

Fixes http://teampulse.telerik.com/view#item/297931


teobugslayer:
:+1:

Fatme:
:+1:

endarova:
:+1:

PR #1185, opened 2015-07-29
Do not fail when build property is not defined

rosen-vladimirov:
When some not mandatory property (like WP8SupportedResolutions) is not defined in .abproject, json schema validation is correct, but we have additional check for our build properties which is preventing the build. Show warning instead of error in such case as the json validation will fail long before that if mandatory property has value undefined.

Fixes http://teampulse.telerik.com/view#item/298156


Fatme:
:+1:

teobugslayer:
:+1:

PR #1186, opened 2015-07-29
Fix finding of Google Play certificate

rosen-vladimirov:
When building for Android AppStore, Google requires min endDate of the certificate to be 23 October 2033. When you have certificate expiring exactly at this date, CLI does not use it and tells you that it cannot find applicable store certificate.
The problem is that our constant for MIN expiration date of the certificate is created with Date only. This creates new Date object with specified date, but the hours in the object depend on the current timezone, so for BG it is set to 3:00:00. So when we validate the certificates expiring at 00:00:00 on 23 October 2033, we do not use them as they are not matching the MIN expiration date.

Fixes http://teampulse.telerik.com/view#item/298197


Fatme:
:+1:  after green

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #1190, opened 2015-08-03
Modify test infrastrucure

rosen-vladimirov:
As common lib is a separate npm package, it has its own tests, that must be run when executing `grunt test` or `npm test` in icenium-cli. Modify test command in package.json in order to run the tests. Change Gruntfile.js in order to ignore/build the correct typescript files in common lib.
Add Bluebird as dependency as it is required for common lib. Update mocha version. Remove `chai` and `mocha` definitions as they are part of common lib now.
Remove yok tests as they are part of common lib now.

Part of http://teampulse.telerik.com/view#item/294238 and http://teampulse.telerik.com/view#item/294240


package.json
```diff
@@ -51,6 +52,7 @@
     "marked-terminal": "1.4.0",
     "minimatch": "2.0.4",
     "mkdirp": "0.3.5",
+    "mocha": "2.2.5",
```
teobugslayer: mocha is both in dependencies and devDependencies with different versions. Which one is correct?

rosen-vladimirov: 2.2.5, I've fixed this :)

rosen-vladimirov:
run ci

Gruntfile.js
```diff
@@ -36,6 +36,16 @@ module.exports = function(grunt) {
 		dateString: now.substr(0, now.indexOf("T")),
 
 		pkg: grunt.file.readJSON("package.json"),
+		commonDtsFileCreation: grunt.file.write(require("path").join("lib", "common", ".d.ts"), ""),
```
teobugslayer: I think you can just do `"lib/common/.d.ts"` - it should work on Windows as well.

Gruntfile.js
```diff
@@ -36,6 +36,16 @@ module.exports = function(grunt) {
...
+		
+		// When there are node_modules inside lib\common directory, CLI behaves incorrectly, so delete this dir.
+		removeCommonNodeModules: require("rimraf")(require("path").join("lib", "common", "node_modules"), function(err) {
```
teobugslayer: You can either concat the path to a single string, or require path and use it everywhere.

Gruntfile.js
rosen-vladimirov: I prefer to use path.join

teobugslayer:
:+1:

PR #1191, opened 2015-08-04
Fix livesync --watch command

Fatme:
http://teampulse.telerik.com/view#item/298483
Should be merged after this PR https://github.com/telerik/mobile-cli-lib/pull/412


rosen-vladimirov:
:+1:

PR #1195, opened 2015-08-10
Refactor remote command

teobugslayer:
Fix bad command line - "--timeout null"
Fix bad command line - lunching from Graphite

Use ES6 imports
Added missing dependency in package.json - shelljs

Fixes http://teampulse.telerik.com/view#item/298968

### DIFF
--- a/lib/services/remote-service.ts
+++ b/lib/services/remote-service.ts
@@ -1,0 +1,82 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import { Request, Response } from "express";
+import * as path from "path";
+import * as os from "os";
+import * as minimatch from "minimatch";
+import * as ip from "ip";
+
+export class RemoteService implements IRemoteService {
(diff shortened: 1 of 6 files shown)


lib/services/remote-service.ts
```diff
@@ -0,0 +1,82 @@
...
+			});
+
+			req.pipe(archive);
+			this.$fs.futureFromEvent(archive, 'finish').wait();
+
+			this.$fs.unzip(this.packageLocation, this.appBuilderDir).wait();
+
+			let appLocation = path.join(this.appBuilderDir, this.$fs.readDirectory(this.appBuilderDir).wait().filter(minimatch.filter("*.app"))[0]);
+
+			this.$iOSEmulatorServices.checkAvailability(false).wait();
```
rosen-vladimirov: isn't it too late to check for iOS simulator availability here?

rosen-vladimirov:
:+1:

PR #1196, opened 2015-08-10
Add shelljs dependency

Mitko-Kerezov:
Used in iOSEmulatorServices


teobugslayer:
No need to - another merge already added this dependency.

PR #1198, opened 2015-08-11
Add support for iojs

rosen-vladimirov:
Change our "native" dependencies to point to newly created tags, where support for iojs 2.x has been added.
http://teampulse.telerik.com/view#item/297772

With latest fibers changes, we can no longer use `future.throw(<some string value>)` as this is no longer supported. Throw real errors instead of strings.


teobugslayer:
:+1: Thy shall not throw a string EVA!

PR #1202, opened 2015-08-14
Fix livesync --companion for ns projects

Fatme:
Fixes http://teampulse.telerik.com/view#item/299070

### DIFF
--- a/lib/providers/device-app-data-provider.ts
+++ b/lib/providers/device-app-data-provider.ts
@@ -75,11 +75,11 @@ export class AndroidNativeScriptCompanionAppIdentifier extends deviceAppDataBase
 	}
 
 	public get deviceProjectRootPath(): string {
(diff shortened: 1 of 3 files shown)


lib/providers/device-app-data-provider.ts
```diff
@@ -76,11 +76,11 @@ export class AndroidNativeScriptCompanionAppIdentifier extends deviceAppDataBase
 	}
 	
 	public get deviceProjectRootPath(): string {
-		return this.getDeviceProjectRootPath(path.join("/data/data", this.appIdentifier, "files"));
+		return this.getDeviceProjectRootPath(path.join("/mnt/sdcard/Android/data", this.appIdentifier, "files"));
```
teobugslayer: Similar path exists in a livesync-service.ts (DEVICE_TMP_DIR_FORMAT_V3) - you may consider extracting a common constant.

Mitko-Kerezov:
:+1:

PR #1209, opened 2015-08-24
Releasing AB 2.11.3

ikoevska:
-Fixed release date
-Dropped the list of features and added link to the official docs to decrease the size of the readme so that it gets published correctly.

NOTE: This is 2.11.3 since we released a 2.11.2 around August 10 to fix the VSE issues.


ikoevska:
Got a formal :+1: from @Mitko-Kerezov

PR #1214, opened 2015-09-04
Generate QR codes only for application packages

mbektchiev:
The server will start returning multiple build result files. The app package is marked with disposition 'BuildResult', while additional configuration files with 'AdditionalFile'.
We need to download the additional files when --download option is specified, but should not generate a QR code for them.

ping @Fatme @rosen-vladimirov @teobugslayer 

refs https://github.com/Icenium/Ice/pull/4204, http://teampulse.telerik.com/view#item/300733

### DIFF
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -42,6 +42,8 @@ declare module Server {
 		solutionPath: string;
 		relativePath: string;
 		localFile?: string;
+		disposition: string;
+		format: string;
 	}
 
 	interface IBuildResult {
--- a/lib/services/build.ts
+++ b/lib/services/build.ts
@@ -81,7 +81,9 @@ export class BuildService implements Project.IBuildService {
 					platform: buildResult.Platform,
 					solution: solutionName,
 					solutionPath: solutionPath,
-					relativePath: buildResult.FullPath
+					relativePath: buildResult.FullPath,
+					disposition: buildResult.Disposition,
+					format: buildResult.Format
 				};
 			});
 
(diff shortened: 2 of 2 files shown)


lib/declarations.d.ts
```diff
@@ -42,6 +42,8 @@ declare module Server {
 		solutionPath: string;
 		relativePath: string;
 		localFile?: string;
+		disposition: string;
```
rosen-vladimirov: as the disposition is not mandatory, maybe we should declare it with `?`

Fatme:
:+1:

lib/declarations.d.ts
rosen-vladimirov: but we are always setting it, so I'm not sure if it is correct to have `?`

rosen-vladimirov:
:+1:

lib/declarations.d.ts
mbektchiev: I'm not sure what is the implication if it is declared with or without `?`. Only older versions of the server will never return it. When I merge the PR in Ice it will always be there. How do you think it would be better - to leave it as required or not?

rosen-vladimirov: well, even if the server does not return it, we'll set it to undefined [here](https://github.com/Icenium/icenium-cli/pull/1214/files#diff-5506b20d356bfab833a3f2bb5fca8bc5R85)
So, let's keep it as is. Ignore my comments above :)

mbektchiev:
run ci

PR #1218, opened 2015-09-09
Fix provision autoselection upon build

Mitko-Kerezov:
Respect the settings object instead of hardcoding AdHoc type provisions.
Kudos for @galexandrov for discovering the bug.


Mitko-Kerezov:
run ci

teobugslayer:
:+1:

PR #1219, opened 2015-09-09
Fix multipart upload

rosen-vladimirov:
Fix the multipart upload where in some cases we produce last chunk with size larger than 20MB. The server returns error for such chunks.


Mitko-Kerezov:
Seems legit :+1:

teobugslayer:
:+1:

PR #1220, opened 2015-09-10
Fixed sys reqs until we introduce support for Node 4

ikoevska:
We need to have this in the master as this is our landing spot in GitHub. We need to fix this readme as soon as we introduce Node 4.x support.


teobugslayer:
:+1:

PR #1221, opened 2015-09-11
Parse correctly build results

teobugslayer:
After @mbektchiev's change the build returns more than one package def. Use the correct IPackageDef to get the built binary package

Fixes http://teampulse.telerik.com/view#item/300909

### DIFF
--- a/lib/commands/deploy.ts
+++ b/lib/commands/deploy.ts
@@ -13,8 +13,7 @@ export class DeployHelper implements IDeployHelper {
 		protected $liveSyncService: ILiveSyncService,
 		protected $errors: IErrors,
 		private $mobileHelper: Mobile.IMobileHelper,
-		private $options: IOptions,
-		private $devicePlatformsConstants:Mobile.IDevicePlatformsConstants) { }
+		private $options: IOptions) { }
 
 	public deploy(platform?: string): IFuture<void> {
 		this.$project.ensureProject();
@@ -26,16 +25,15 @@ export class DeployHelper implements IDeployHelper {
 			this.$errors.failWithoutHelp("On your current OS, you cannot deploy apps on connected %s devices.", this.$mobileHelper.normalizePlatformName(platform));
 		}
 
+		if (this.$options.companion) {
+			return this.$liveSyncService.livesync(platform);
+		}
+
 		return this.deployCore(platform);
 	}
 
 	private deployCore(platform: string): IFuture<void> {
 		return ((): void => {
-			if (this.$options.companion) {
-				this.$liveSyncService.livesync(platform).wait();
-				return;
-			}
-
 			this.$devicesServices.initialize({ platform: platform, deviceId: this.$options.device}).wait();
 			let packageName = this.$project.projectData.AppIdentifier;
 			let packageFile: string = null;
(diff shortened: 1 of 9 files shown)


lib/services/extensions-service-base.ts
```diff
@@ -52,7 +52,7 @@ export class ExtensionsServiceBase {
 
 			if( this.shouldUpdatePackage(cachedVersion, extensionVersion) ) {
 				this.$logger.printInfoMessageOnSameLine(`Updating ${packageName} package...`);
-				let zipFileName = temp.path({ path:  path.join(this.cacheDir, packageName + ".zip") });
+				let zipFileName = path.join(this.cacheDir, packageName + ".zip");
```
Mitko-Kerezov: Due to `temp` module not being used anymore, there is a chance that this file is not deleted if an exception is raised during unzip for example

Mitko-Kerezov: Also the `temp` module is not used anymore in this file, so the `import` statement in its beginning becomes obsolete

lib/commands/deploy.ts
```diff
@@ -44,16 +43,17 @@ export class DeployHelper implements IDeployHelper {
...
+					if (helpers.versionCompare(deviceVersion, deploymentTarget) < 0) {
 						this.$logger.error(`You cannot deploy on device ${device.deviceInfo.identifier} with OS version ${deviceVersion} when iOSDeploymentTarget is set to ${deploymentTarget}.`);
 						return Future.fromResult();
 					}
 				}
 
-				if(!packageFile) {
+				if (!packageFile) {
 					let packageDefs = this.$buildService.deploy(this.$devicesServices.platform, device).wait();
+					packageDefs = _.filter(packageDefs, (def: Server.IPackageDef) => !def.disposition || def.disposition === "BuildResult");
```
Mitko-Kerezov: Consider extracting this logic in a separate method (in helpers for example) because it seems to be repeated on multiple places

lib/services/extensions-service-base.ts
teobugslayer: The temp module have never worked. We used wrong syntax and it created a temporary file which we never used.

lib/commands/deploy.ts
teobugslayer: I really could not name this method nor did I found its proper place :( `getBinaryPackage`?

lib/commands/emulate.ts
```diff
@@ -68,6 +68,7 @@ export class EmulateIosCommand implements ICommand {
 					downloadedFilePath: path.join(tempDir, "package.ipa"),
 					buildForiOSSimulator: true
 				}).wait();
+				packageDefs = _.filter(packageDefs, (def: Server.IPackageDef) => !def.disposition || def.disposition === "BuildResult");
```
rosen-vladimirov: you can use build for deploy here and call 
```TypeScript
this.$fs.unzip(packageDefs, tempDir).wait()
```

lib/commands/deploy.ts
```diff
@@ -44,17 +43,16 @@ export class DeployHelper implements IDeployHelper {
 
 			let action = (device: Mobile.IDevice): IFuture<void> => {
 				let deploymentTarget = this.$project.projectData.iOSDeploymentTarget;
-				if(deploymentTarget && device.deviceInfo.platform.toLowerCase() === this.$devicePlatformsConstants.iOS.toLowerCase()) {
+				if (deploymentTarget && device.deviceInfo.platform.toLowerCase() === this.$devicePlatformsConstants.iOS.toLowerCase()) {
```
rosen-vladimirov: you can use 
```TypeScript
if(deploymentTarget && this.$mobileHelper.isiOSPlatform(device.deviceInfo.platform))
```

rosen-vladimirov:
:+1:  after green build

PR #1224, opened 2015-09-14
Add support for AppManager LiveSync mandatory app updates to CLI

manoldonev:
http://teampulse.telerik.com/view#item/301163

### DIFF
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -34,7 +34,7 @@ export class Options extends commonOptionsLibPath.OptionsBase {
 			sendPush: { type: OptionType.Boolean },
 			sendEmail: { type: OptionType.Boolean },
 			group: { type: OptionType.Array },
-			isMandatory: { type: OptionType.Boolean},
+			mandatory: { type: OptionType.Boolean},
 			default: {type: OptionType.Boolean},
 			var: {type: OptionType.Object},
 			answers: { type: OptionType.String },
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -34,6 +34,7 @@ export class Options extends commonOptionsLibPath.OptionsBase {
 			sendPush: { type: OptionType.Boolean },
 			sendEmail: { type: OptionType.Boolean },
 			group: { type: OptionType.Array },
+			isMandatory: { type: OptionType.Boolean},
 			default: {type: OptionType.Boolean},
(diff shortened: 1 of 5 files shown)


manoldonev:
@teobugslayer

lib/options.ts
```diff
@@ -34,6 +34,7 @@ export class Options extends commonOptionsLibPath.OptionsBase {
 			sendPush: { type: OptionType.Boolean },
 			sendEmail: { type: OptionType.Boolean },
 			group: { type: OptionType.Array },
+			isMandatory: { type: OptionType.Boolean},
```
ligaz: Do we need to be prefixed `is` or it can be just `mandatory`? For example `sendPush` is not `shouldSendPush`.

lib/services/appmanager-service.ts
```diff
@@ -138,7 +138,10 @@ class AppManagerService implements IAppManagerService {
 
 			this.$project.importProject().wait();
 			this.$logger.printInfoMessageOnSameLine("Publishing patch for " + platforms.join(", ") + "...");
-			this.$progressIndicator.showProgressIndicator(this.$server.tam.uploadPatch(this.$project.projectData.ProjectName, this.$project.projectData.ProjectName, <any>{ Platforms: platforms }), 2000).wait();
+
+			let patchData = <any> { Platforms: platforms, IsMandatory: this.$options.isMandatory };
```
ligaz: What's the need for the `any` cast?

lib/options.ts
manoldonev: Yup, you are right -- updated it for consistency.

lib/services/appmanager-service.ts
manoldonev: The any cast was present in the original implementation -- my understanding is that it was used in order to make the { Platforms: platforms } object assignable to Server.PatchData. As Server.PatchData.Platforms is of type Server.DevicePlatform[] while the platforms argument in this method is of type string[].

I personally could not find a better and not too verbose way of converting string[] to Server.DevicePlatform[] so I kept the any cast. One option is to change Server.PatchData.Platforms type to string[], then there will be no need for any cast (either way some of the api now use the platform concept as string, and other parts -- as Server.DevicePlatform const enum.

teobugslayer:
:+1:

PR #1227, opened 2015-09-17
Fix Livesync service

rosen-vladimirov:
Remove a parameter that is no longer used in common's master branch from
livesync service.


teobugslayer:
Sorry that forgot to remove it. :+1: from me

PR #1239, opened 2015-09-28
Make npmsearch stricter

rosen-vladimirov:
Add quotes around name when searching the npm - this will make the search stricter. For example when registry-validator is checked, npmsearch api will find everything that has validator in it in case the quotes are missing.


ndinev:
:+1:

Mitko-Kerezov:
:+1:

PR #1240, opened 2015-09-28
Use intersection for plugin|find command

rosen-vladimirov:
Plugin find command must show intersection of all provided parameters.
Also use encodeUriComponent for all parameters that user will fill when calling npmsearch.
Add error message when we cannot find any plugin matching provided arguments.

### DIFF
--- a/lib/services/nativescript-project-plugins-service.ts
+++ b/lib/services/nativescript-project-plugins-service.ts
@@ -8,6 +8,7 @@ import * as shelljs from "shelljs";
 import * as semver from "semver";
 import {getFuturesResults} from "../common/helpers";
 import {MarketplacePluginData} from "../plugins-data";
+import Future = require("fibers/future");
 import temp = require("temp");
 temp.track();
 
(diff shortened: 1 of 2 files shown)


lib/services/nativescript-project-plugins-service.ts
```diff
@@ -48,7 +49,16 @@ export class NativeScriptProjectPluginsService implements IPluginsService {
...
-			return getFuturesResults<IBasicPluginInformation>(findPluginByNameFutures, plugins => !!plugins && !!plugins.length);
+			Future.wait(findPluginByNameFutures);
+			let findPluginsResults: IBasicPluginInformation[][] = findPluginByNameFutures.map(f => f.get()).filter(plugins => !!plugins && !!plugins.length);
+
+			let pluginsNamesInGroups = findPluginsResults.map(res => _.map(res, r => r.name));
+			let matchingNames = _.intersection.apply(this, pluginsNamesInGroups);
+
+			return _(findPluginsResults)
+					.flatten<IBasicPluginInformation>()
+					.filter(basicInfo => _.contains(matchingNames, basicInfo.name))
```
rosen-vladimirov: this way each plugin, that is found in both collections, will be shown twice, show only one instance of a plugin

teobugslayer:
:+1:

lib/services/nativescript-project-plugins-service.ts
```diff
@@ -311,7 +321,7 @@ export class NativeScriptProjectPluginsService implements IPluginsService {
 		return ((): IPlugin => {
 			let plugin: IPlugin;
 			try {
-				let url = `${NativeScriptProjectPluginsService.NPM_REGISTRY_URL}/${packageName}?version=${version || "latest"}`;
+				let url = `${NativeScriptProjectPluginsService.NPM_REGISTRY_URL}/${encodeURIComponent(packageName)}?version=${encodeURIComponent(version || "latest")}`;
```
Fatme: I suggest you to extract this to separate function
``` 
private buildUrl(packageName: string, version: string): string {
   return `${NativeScriptProjectPluginsService.NPM_REGISTRY_URL}/${encodeURIComponent(packageName)}?version=${encodeURIComponent(version)}`;
}

File: lib/services/nativescript-project-plugins-service.ts [hunk 3]
```diff
@@ -48,7 +49,16 @@ export class NativeScriptProjectPluginsService implements IPluginsService {
 	public findPlugins(keywords: string[]): IFuture<IBasicPluginInformation[]> {
 		return ((): IBasicPluginInformation[] => {
 			let findPluginByNameFutures = _.map(keywords, keyword => this.findPluginsByName(keyword));
-			return getFuturesResults<IBasicPluginInformation>(findPluginByNameFutures, plugins => !!plugins && !!plugins.length);
+			Future.wait(findPluginByNameFutures);
+			let findPluginsResults: IBasicPluginInformation[][] = findPluginByNameFutures.map(f => f.get()).filter(plugins => !!plugins && !!plugins.length);
+
+			let pluginsNamesInGroups = findPluginsResults.map(res => _.map(res, r => r.name));
+			let matchingNames = _.intersection.apply(this, pluginsNamesInGroups);
```
Fatme: Do you need to use `.apply`? Can you just call .intersection?

rosen-vladimirov: nope, `intersection` works with many arrays passed as separate parameters, pluginsNamesInGroups is array of arrays. That's what the apply is doing for me - unpacking the array of arrays to different arguments, each of them is array. This way intersection works correctly

ndinev:
:+1:

Fatme:
:+1:

PR #1243, opened 2015-09-30
Ignore npm scripts of {N} plugins

rosen-vladimirov:
Ignore npm scripts when installing {N} plugins/npm modules to temp
directory as they may cause unexpected behavior or client's machines.
Fixes:
```
appbuilder plugin add https://github.com/rosen-vladimirov/NativeScriptPluginTest/tarball/master
```
which throws `ENOENT error`.


ndinev:
:+1: this is a blocking problem for adding NS plugin from URL

Mitko-Kerezov:
:+1:

Fatme:
:+1:

PR #1246, opened 2015-10-01
Use correct watchGlob

Fatme:
In order to work correctly for added files we've changed the usage of gaze module - https://github.com/telerik/mobile-cli-lib/blob/master/services/usb-livesync-service-base.ts#L106. This requires to change the passed watch glob from cli.

If livesync --watch command is executed only the first sync works correctly. After that the changes aren't reported from the watch.


teobugslayer:
:+1:

Mitko-Kerezov:
:+1:

PR #1251, opened 2015-10-07
Print call stack when log trace/debug is used

rosen-vladimirov:
When `--log trace` or `--log debug` is used, we should print call stack of failure.


teobugslayer:
:+1:

PR #1257, opened 2015-10-19
Print information about local plugins

rosen-vladimirov:
When plugin in referenced via local path or github url, `appbuilder plugin` command
does not show information about it.
Read the package.json of the local plugin in order to get data for it.
Install the plugin referenced with github url to temp dir and read its package.json.

Fixes http://teampulse.telerik.com/view#item/303583

### DIFF
--- a/lib/services/nativescript-project-plugins-service.ts
+++ b/lib/services/nativescript-project-plugins-service.ts
@@ -73,7 +73,13 @@ export class NativeScriptProjectPluginsService implements IPluginsService {
 			if(content && content.dependencies) {
 				let items =  _.map(content.dependencies, (version: string, name: string) => {
 					let marketplacePlugin = _.find(this.getMarketplacePlugins().wait(), pl => pl.data.Name === name && pl.data.Version === version);
-					return marketplacePlugin || this.getDataForNpmPackage(name, version).wait();
+					let plugin = marketplacePlugin || this.getDataForNpmPackage(name, version).wait()
+								|| this.getDataForLocalPlugin(name, version).wait()
+								|| this.getDataFromGitHubUrl(name, version).wait();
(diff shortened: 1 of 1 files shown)


teobugslayer:
```
does not prompt for Cordova upgrade to 3.7.0 when WPSdk is upgraded to 8.1 and FrameworkVersion is already 3.7.0 (550ms)
09:55:42     Init command tests
09:55:42       NativeScript project
09:55:42 AssertionError: NativeScript Blank template does not contain mandatory 'tns_modules' directory. This directory is required in init command. You should check if this is problem with the template or change init command to use another file.: expected false to be true
```

rosen-vladimirov:
I'll fix the tests in separate PR and rebase this one after merging it :)

rosen-vladimirov:
I've rebased this PR, so it should be fine now :)

lib/services/nativescript-project-plugins-service.ts
```diff
@@ -358,6 +343,68 @@ export class NativeScriptProjectPluginsService implements IPluginsService {
...
+				pathToPlugin = pathToPlugin.replace("file:", "");
+			}
+
+			if(this.checkIsValidLocalPlugin(pathToPlugin).wait()) {
+				let fullPath = path.resolve(pathToPlugin);
+				let packageJsonContent = this.$fs.readText(path.join(fullPath, this.$projectConstants.PACKAGE_JSON_NAME)).wait();
+				plugin = this.constructNativeScriptPluginData(packageJsonContent).wait();
+			}
+
+			return plugin;
```
teobugslayer: This returns implicitly `undefined`. Should we return an explicit `null` instead?

teobugslayer:
Looks good, :+1: but check my remark.

PR #1258, opened 2015-10-19
Fix init command for NativeScript projects

rosen-vladimirov:
Currently `appbuilder init` command checks that the project has `app/tns_modules` dir.
We do not have such directory anymore, so check the `package.json` instead of it.
Fix unit tests.


Mitko-Kerezov:
:+1:

PR #1263, opened 2015-10-21
Unify mobileframework and prop <subcommand> FrameworkVersion commands

rosen-vladimirov:
`$ appbuilder mobileframework` should work in the same way as `$ appbuilder prop print FrameworkVersion`
Same is valid for `$ appbuilder mobileframework set <version>` and `$ appbuilder prop set FrameworkVersion <version>`.
In order to achieve this, register new commands - `prop|print|frameworkversion` and `prop|set|frameworkversion`.


teobugslayer:
:+1: but we may need some more unit tests

teobugslayer:
:+1:

PR #1265, opened 2015-10-23
Fix NativeScript project unit tests

rosen-vladimirov:
Fix unit tests that create {N} project to expect 1.4.3 version.


Fatme:
:+1:  after green build

PR #1266, opened 2015-10-23
Fix starting of iOS Simulator

teobugslayer:
Fix starting of iOS Simulator

Fixes http://teampulse.telerik.com/view#item/300209


rosen-vladimirov:
:+1:  after passing unit tests

Fatme:
:+1:

PR #1267, opened 2015-10-26
Delete styles.xml when migrating to {N} 1.4.3+

rosen-vladimirov:
Due to bug in {N} 0.4.2 in our templates we have `AppResources/Android/values/styles.xml`
and `AppResources/Android/values-v21/styles.xml` files. When migrating to 1.4.3 or later,
we must remove these files as they are change in {N} android runtime.
The real styles will be taken during build.

Add unit test for renameIfExists method of fs.

http://teampulse.telerik.com/view#item/303971


teobugslayer:
The QA tests currently expect styles.xml, so we need to merge with red tests :(

Mitko-Kerezov:
:+1:

PR #1259, opened 2015-10-19
Track user's edition type by using tenant.editionType

rosen-vladimirov:
Track user's edition type by using new property - editionType.
Implements http://teampulse.telerik.com/view#item/303569

Should be merged after https://github.com/Icenium/Ice/pull/4436

PR #1271, opened 2015-10-28
Track user's edition type by using tenant.editionType

rosen-vladimirov:
Track user's edition type by using new property - editionType.
Implements http://teampulse.telerik.com/view#item/303569


rosen-vladimirov:
Just retargeting to release, change is already approved [here](https://github.com/Icenium/icenium-cli/pull/1259)

PR #1278, opened 2015-10-29
Support npm 3+ when adding nativescript plugins

rosen-vladimirov:
In nativescript project, when a plugin is added, CLI's trying to install it in local dir
in case you add it via github url or local dir.
The code is trying to detect the installation directory based on the output.
But with `npm 3+` the output is changed a lot, so we have to check the installation dir and see if there is node_modules dir there.
If we have such dir with only one package, this must be our package, so let's use it.


teobugslayer:
:+1:

PR #1281, opened 2015-11-02
Integrate proton branch

rosen-vladimirov:
Renamed logcatPrinter to DeviceLogProvider. This provider is now used for both Android and iOS device logs.
Integrate all changes from proton branch.


lib/providers/device-log-provider.ts
```diff
@@ -0,0 +1,39 @@
...
+"use strict";
+
+export class DeviceLogProvider implements Mobile.IDeviceLogProvider {
+	private static LINE_REGEX = /.\/(.+?)\s*\(\s*(\d+?)\): (.*)/;
+
+	constructor(private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
+				private $logger: ILogger) { }
+
+	public logData(lineText: string, platform: string, deviceIdentifier: string): void {
+		if(!platform || platform === this.$devicePlatformsConstants.iOS) {
```
teobugslayer: This is case-sensitive comparison for platform name, is this OK?

lib/providers/device-log-provider.ts
```diff
@@ -0,0 +1,39 @@
...
+				} else {
+					this.$logger.out(log.message);
+				}
+			}
+		}
+	}
+
+	private getConsoleLogFromLine(lineText: String): any {
+		let acceptedTags = ["chromium", "Web Console", "JS"];
+		//sample line is "I/Web Console(    4438): Received Event: deviceready at file:///storage/emulated/0/Icenium/com.telerik.TestApp/js/index.js:48"
```
teobugslayer: I think this comment would be way more useful if moved near the LINE_REGEX definition.

teobugslayer:
:+1:

PR #1286, opened 2015-11-04
Fix appbuilder hooks

rosen-vladimirov:
Due to changes in hooks service, our current code is trying to execute both `.ts` and `.js` hook files.
As the `.ts` do not contain any TypeScript specific logic, just remove them and use the `.js` directly.


Fatme:
:+1:

PR #1291, opened 2015-11-10
Introduce SysInfo class

rosen-vladimirov:
Introduce SysInfo class that extends SysInfoBase from common lib.
This class holds AppBuilder specific implementation of gathering SysInfo.


Mitko-Kerezov:
:+1:

teobugslayer:
:+1:

PR #1292, opened 2015-11-11
Allow export of apps and solutions using new server API

rosen-vladimirov:
Due to changes in Server API, we have to allow exporting of apps and solutions.
In order to get projects from apps, we should call TAP endpoint.
For not migrated projects we should call the old server endpoint.
The export procedure requires calling same endpoint for migrated and not migrated projects.
When exporting old projects (from solutions), we should pass the solution name as appId and send the magic Header in the request.
When exporting new projects (from apps), we should pass real appId and DO NOT send the magic Header.
In order to achieve this, introduce new class - ServiceProxyBase and use it for TAP requests.
Introduce AppBuilderServiceProxy class that extends ServiceProxyBase and register it as "serviceProxy", so we'll keep current behavior for server calls.
Add new method in AppBuilderServiceProxy that allows sending server requests with passing "X-Icenium-SolutionSpace" header.
Add "NOT MIGRATED" string to solutions when they are listed.

### DIFF
--- a/lib/commands/cloud-projects.ts
+++ b/lib/commands/cloud-projects.ts
@@ -10,8 +10,8 @@ class SolutionIdCommandParameter implements ICommandParameter {
 	validate(validationValue?: string): IFuture<boolean> {
 		return (() => {
 			if(validationValue) {
-				this.$remoteProjectService.getSolutionName(validationValue).wait();
-				return true;
+				let app = this.$remoteProjectService.getSolutionData(validationValue).wait();
+				return !!app;
 			}
 
 			return false;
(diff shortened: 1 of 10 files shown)


lib/commands/cloud-projects.ts
```diff
@@ -42,16 +42,15 @@ export class CloudListProjectsCommand implements ICommand {
...
 	execute(args: string[]): IFuture<void> {
 		return (() => {
-			let slnName: string;
-			if(args[0]) {
-				slnName = this.$remoteProjectService.getSolutionName(args[0]).wait();
-			} else {
-				let solutions = this.$remoteProjectService.getSolutions().wait().map(sln => sln.name);
-				if (this.$options.all || !commonHelpers.isInteractive()) {
-					this.printList(solutions);
+			let slnName = args[0];
```
teobugslayer: You can move this line down, to be near the place it is used.

lib/declarations.d.ts
```diff
@@ -1106,3 +1108,57 @@ interface INativeScriptResources {
...
+ * Describes information for applications returned by TAP
+ */
+interface ITapAppData {
+	/**
+	 * AccountId of the owner of the project.
+	 */
+	accountId: string;
+
+	/**
+	 * Application Identifier.
```
teobugslayer: It is "platform application" identifier.

teobugslayer:
:+1: after green build

rosen-vladimirov:
run ci

PR #1296, opened 2015-11-13
Fix livesync

Mitko-Kerezov:
Includes
+ Fixing for QR code generation for {N} Ion
+ Fix for cable LiveSync for both Cordova and {N}

After this commit every kind of LiveSync works as expected with the following limitation. **When one deploys code directly to any companion app without first scanning a QR code no subsequent cloud LiveSync-ing will work in said companion app.** This limitation comes from the fact that whenever cable LiveSync-ing to a companion app the companion app is only told to get the files and refresh - no livesync url is provided. In order to obtain a LiveSync url on the other hand, one must scan a QR code.

TP Refs: 
+ [Build android --companion generates an invalid QR code](http://teampulse.telerik.com/view#item/304910)
+ [Impossible to cloud livesync if initially cable livesynced](http://teampulse.telerik.com/view#item/305176) <- while this bug is not fixed the behavior between CLI and VSE is balanced out with this fix

Ping @Fatme @rosen-vladimirov @teobugslayer

### DIFF
--- a/lib/providers/device-app-data-provider.ts
+++ b/lib/providers/device-app-data-provider.ts
@@ -81,7 +81,7 @@ export class AndroidNativeScriptCompanionAppIdentifier extends deviceAppDataBase
 	}
 
 	public get liveSyncFormat(): string {
-		return null;
+		return "%s/Mist/MobilePackage/nsredirect?token=%s";
 	}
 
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
:+1:  after green build

PR #1297, opened 2015-11-13
Document --justlaunch

teobugslayer:
Fixes http://teampulse.telerik.com/view#item/300209


rosen-vladimirov:
:+1:  but I prefer to remove it from common and work with this option in NativeScript CLI
This requires a lot of refactoring, so documenting it for the moment is the best solution :)

PR #1302, opened 2015-11-16
Remove support for Mobile Websites

teobugslayer:
`cloud export` should continue working.
This PR should be merged post-3.0 release after Web Sites support is toggled off for all users.


lib/services/samples-service.ts
```diff
@@ -46,6 +46,7 @@ export class SamplesService implements ISamplesService {
 		return (() => {
 			this.$logger.info("You can choose a sample from the following: %s", EOL);
 			if(framework) {
+				if (framework)
```
rosen-vladimirov: hm, I'm pretty sure one if is enough for this check :smile_cat:

rosen-vladimirov:
`dynamicHelpProvider` defines `isMobileWebsite` variable for our help, I believe we should remove it as well

teobugslayer:
We need to leave isMobileWebsite  defined. We still support `cloud export` and `publish` commands and their help must be showable (if that is a word).

teobugslayer:
run ci

lib/project/web-site-project.ts
```diff
@@ -22,7 +22,7 @@ export class MobileWebSiteProject extends frameworkProjectBaseLib.FrameworkProje
 	}
 
 	public get name(): string {
-		return this.$projectConstants.TARGET_FRAMEWORK_IDENTIFIERS.MobileWebsite;
+		throw new Error("Тhe development of mobile websites is discontinued. For more information, see http://www.telerik.com/blogs/whats-next-for-telerik-platform-and-web-projects");
```
Fatme: `this.$errors.failWithoutHelp` maybe. Also you extract the message to separate variable/function.

teobugslayer:
run ci

teobugslayer:
Will retarteg against master

PR #1303, opened 2015-11-18
Do not show NOT MIGRATED string when user cannot migrate

rosen-vladimirov:
In case the user is not allowed to migrate his projects to apps, we shouldn't show NOT MIGRATED string when listing them.
Fixes http://teampulse.telerik.com/view#item/305450


teobugslayer:
:+1:

PR #1304, opened 2015-11-19
Fix solution export by index

rosen-vladimirov:
When export solution by index is called, we use helpers.findByNameOrIndex method.
This is not correct, as the method is sorting the the provided collection, so if index is used, it is not the one that user specified.
Fixes http://teampulse.telerik.com/view#item/305538


teobugslayer:
:+1:

PR #1306, opened 2015-11-24
Fix support for node 5.1.0 by updating plugman

rosen-vladimirov:
Fix support for node 5.1.0. Curently the installation is failling because we are using old plugman.
This old plugman depends on very old version of npm. During installation with npm 3.3.11 or later, the plugman fails.
This is a bug in npm itself (https://github.com/npm/npm/issues/10234). However updating our version of plugman is fixing the installation.
The new plugman had change the callback parameters - previously it was one parameter, now it has two - error and result.
Fix our callbacks to have two arguments.


teobugslayer:
:+1:

Mitko-Kerezov:
:+1:

tytodorov:
:+1:

PR #1310, opened 2015-11-27
Update sys-info according to latest common changes

rosen-vladimirov:
Update sys-info in order to use CLI's package.json.

### DIFF
--- a/lib/services/livesync-service.ts
+++ b/lib/services/livesync-service.ts
@@ -84,7 +84,7 @@ export class LiveSyncService extends usbLivesyncServiceBaseLib.UsbLiveSyncServic
 				"ios": IOSLiveSyncService
 			};
 
-			let livesyncData=  {
+			let livesyncData = {
 				platform: platform,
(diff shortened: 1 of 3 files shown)


lib/services/livesync-service.ts
```diff
@@ -92,7 +92,8 @@ export class LiveSyncService extends usbLivesyncServiceBaseLib.UsbLiveSyncServic
 				watchGlob: projectDir,
 				platformSpecificLiveSyncServices: platformSpecificLiveSyncServices,
 				notInstalledAppOnDeviceAction: notInstalledAppOnDeviceAction,
-				notRunningiOSSimulatorAction: () => Future.fromResult()
+				notRunningiOSSimulatorAction: () => Future.fromResult(),
+				getApplicationPathForiOSSimulatorAction:() => Future.fromResult("")
```
rosen-vladimirov: Ping @Fatme - I'm really not sure what to pass here, I've added it as the code cannot be transpiled without this.

Fatme:
:+1:

PR #1311, opened 2015-12-02
Start android applications after deploy

rosen-vladimirov:
Fixes http://teampulse.telerik.com/view#item/306407

Add canStartApplication to applicationManager

We have different cases when we can start the application, based on current platform and device platform.
Add `canStartApplication` in applicationManager's which evaluates all circumstances and can be used to check if we are able to start the application on device.


Mitko-Kerezov:
:+1:

teobugslayer:
:+1:

PR #1313, opened 2015-12-03
Allow plugin fetch by identifier

rosen-vladimirov:
Currently plugin fetch is not working when marketplace plugin identifier is passed.
Also it does not work for org.apache.cordova.inappbrowser (looks like the id has been changed, but plugman search still finds the old id, while plugman fetch fails).

Fixes http://teampulse.telerik.com/view#item/306489

### DIFF
--- a/lib/services/cordova-project-plugins-service.ts
+++ b/lib/services/cordova-project-plugins-service.ts
@@ -64,15 +64,35 @@ export class CordovaProjectPluginsService implements IPluginsService {
 				let result = this.$cordovaPluginsService.fetch(pluginIdentifier).wait();
 				this.$logger.out(result);
 			} else {
+				let identifier = this.getPluginBasicInformation(pluginIdentifier).name.toLowerCase();
+				let plugin = _.find(this.getAvailablePlugins(), pl => pl.data.Identifier.toLowerCase() === identifier || pl.data.Name.toLowerCase() === identifier);
+				let pluginUrl: string = plugin && plugin.data && plugin.data.Url ? plugin.data.Url : null;
 				let plugins = this.$cordovaPluginsService.getPlugins([pluginIdentifier]);
 				let pluginKeys = Object.keys(plugins);
 				let pluginsCount = pluginKeys.length;
 				if (pluginsCount === 0) {
-					this.$logger.out("There are 0 matching plugins.");
-				} else if (pluginsCount > 1 && pluginKeys[0] !== pluginIdentifier) {
-					this.$logger.out("There are more then 1 matching plugins.");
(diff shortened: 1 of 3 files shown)


teobugslayer:
A unit test may be a good addition.

rosen-vladimirov:
@teobugslayer I've added some tests, we can add more later, but these verify exactly what this PR fixes :) 
I've also tried to keep current behavior as much as possible.

lib/services/cordova-project-plugins-service.ts
```diff
@@ -64,15 +64,31 @@ export class CordovaProjectPluginsService implements IPluginsService {
...
 			} else {
+				let identifier = this.getPluginBasicInformation(pluginIdentifier).name.toLowerCase();
+				let plugin = _.find(this.getAvailablePlugins(), pl => pl.data.Identifier.toLowerCase() === identifier || pl.data.Name.toLowerCase() === identifier);
+				let pluginUrl: string = plugin && plugin.data && plugin.data.Url ? plugin.data.Url : null;
 				let plugins = this.$cordovaPluginsService.getPlugins([pluginIdentifier]);
 				let pluginKeys = Object.keys(plugins);
 				let pluginsCount = pluginKeys.length;
 				if (pluginsCount === 0) {
-					this.$logger.out("There are 0 matching plugins.");
+					if(pluginUrl) {
```
teobugslayer: This code is starting to become a too complex. On the top of my head, I suggest this change:

```
if (pluginsCount === 0) {
	if(pluginUrl) {
		this.$logger.out(this.$cordovaPluginsService.fetch(pluginUrl).wait());
	} else {
		this.$logger.out("There are 0 matching plugins.");
	}
	return;
}

if (pluginsCount > 1 && pluginKeys[0] !== pluginIdentifier) {
	this.$logger.out(`There are more then 1 matching plugins: ${pluginKeys.join(", ")}.`);
	return;
}

try {
	this.$logger.out(this.$cordovaPluginsService.fetch(pluginKeys[0]).wait());
} catch (err) {
	if(pluginUrl) {
		this.$logger.trace(`Error while trying to fetch plugin with id ${pluginIdentifier} via plugman. Error is: ${err.message}.`);
		this.$logger.out(this.$cordovaPluginsService.fetch(pluginUrl).wait());
	} else {
		this.$errors.fail(err.message);
	}
}
```

Alternatively, move the code from bodies of if and else clauses into dedicated functions.

teobugslayer:
tests are beautiful! :+1: but consider my suggestion to make the code more readable.

teobugslayer:
:+1:

PR #1314, opened 2015-12-03
Add --skipUi option

rosen-vladimirov:
Add --skipUi option and if it is passed, do not open file on `appbuilder edit-configuration <smth>`.


ndinev:
:+1: Needed for testability reasons

teobugslayer:
Either create a help article, or document it with a comment in the code.

rosen-vladimirov:
Do we want help for this? I prefer to keep it internal (comment in the code could be enough).

rosen-vladimirov:
I was thinking about this code. In fact isn't it better to skip the opening in case the console is not interactive (maybe we should handle this is `$opener`). This way we'll not have to add the code for this new option on every single place where we want to skip opening something. Also QAs will not modify their tests at all, they'll just work.

teobugslayer:
In my experience, at least on Windows, running a script does not make the console non-interactive. One has to specifically write code similar to `appbuilder command > nul` to create non-interactive i/o.

rosen-vladimirov:
ok, so let's keep the new option and conisder improving the opener anyway, right?

rosen-vladimirov:
ok, I've documented the option and applied sorting on the two dictionaries (that's why the changes are so much there), but it will be easier for maintain from now on :) 
Ping @teobugslayer

teobugslayer:
:+1:

PR #1315, opened 2015-12-04
Fix livesync/deploy to NS companion

Mitko-Kerezov:
Result command file should be called `telerik.livesync.commands` instead of `temp.commands.file`
Ping @teobugslayer 
TP Reference: [[CLI][NativeScript] Fail to deploy/livesync to NS companion](http://teampulse.telerik.com/view#item/306613)


rosen-vladimirov:
:+1:

PR #1316, opened 2015-12-07
Do not call adb kill-server on uninstall

teobugslayer:
Make shelljs opeartions silent

See https://github.com/NativeScript/nativescript-cli/issues/1247


rosen-vladimirov:
:+1:

PR #1321, opened 2015-12-14
Fix Genymotion emulator

rosen-vladimirov:
Fix Genymotion emulator starting by updating common lib, where a fix is applied for checking if running device is Genymotion emulator.


teobugslayer:
:+1:

PR #1322, opened 2015-12-15
Make sure CLI works correctly with plugins

rosen-vladimirov:
When marketplace plugins and core plugins have same identifier, CLI does not work correctly.
Make sure to use the marketplace plugin whenever there's plugin with the same identifier in integrated plugins and in marketplace.
Make sure to remove all plugin instances when changing the version, for example if plugin is enabled from integrated plugins, and you try to update it, we will use the marketplace plugin.
So we will try to add new instance of the plugin in `.<config>.abproject`, for example:
Before:
```JSON
{
    "CorePlugins": [ "com.telerik.feedback"]
}
```

After `appbuilder plugin add com.telerik.feedback` the file should be:
```JSON
{
    "CorePlugins": [ "com.telerik.feedback@<selected_version>"]
}
```

More details for the issue - http://teampulse.telerik.com/view#item/307168


rosen-vladimirov:
run ci

mbektchiev:
:+1:

PR #1323, opened 2015-12-15
Require logger and update to latest common lib

rosen-vladimirov:
`logger` has been required in common lib, but due to some Proton requirements, it should be required from each CLI.
Fix this by adding it to bootstrap and update to latest common lib.


teobugslayer:
:+1:

PR #1326, opened 2015-12-23
LowerCase the parameter passed to $ dev-config-apply

Mitko-Kerezov:
It can cause problems in case-sensitive OS.
Ping @mbektchiev


mbektchiev:
:+1:

PR #1327, opened 2015-12-29
Use deviceLogProvider from common

rosen-vladimirov:
Use deviceLogProvider from common lib as both icenium-cli and nativescript-cli have the same logic.
This will fix the following issues:
* `appbuilder device log` does not print full logs for Android devices
* logs for Android devices with API Level 23 are not correct. In fact CLI was crashing when trying to print logs of such devices.


rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #1329, opened 2016-01-04
Allow autocompletion of lowercased properties

rosen-vladimirov:
In case user is using only lowercased values for properties, autocompletion is not working.
For example `appbuilder prop print core` is not autocompleted to coreplugins.
In fact the command will work with lowercased values, so enable using them.
Add lowercased values to the ones used for autocompletion.
With this change the above will work, but scenarios like `appbuilder prop print cOre` will not be autocompleted anyway.

Fixes http://teampulse.telerik.com/view#item/292405


teobugslayer:
:+1: after green build

PR #1333, opened 2016-01-13
Use plugin variables for NativeScript plugins

rosen-vladimirov:
Allow usage of plugin variables of NativeScript plugins. Use `--var` option to pass values, for example:
`appbuilder plugin add plugin-var-plugin --var.APP_NAME test --var.FACEBOOK_APP_ID fb_appid`
For MarketPlace plugins we should use the Variables from Server, but currently it does not return correct info about them. So call registry.npmjs.org to get package.json and get variables info from there.
For plugins from npm, get the info from `registry.npmjs.org`
For local plugins get the info from its package.json.
For plugins passed as URLs get the info from the temp dir where it will be installed.

Use defaultValue specified in plugin for prompter's default value or when console is not interactive.

Implements http://teampulse.telerik.com/view#item/308123

### DIFF
--- a/lib/services/nativescript-project-plugins-service.ts
+++ b/lib/services/nativescript-project-plugins-service.ts
@@ -9,6 +9,7 @@ import * as semver from "semver";
 import {getFuturesResults} from "../common/helpers";
 import {MarketplacePluginData} from "../plugins-data";
 import Future = require("fibers/future");
+import { isInteractive } from "../common/helpers";
 import temp = require("temp");
 temp.track();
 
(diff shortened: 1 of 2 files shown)


lib/services/nativescript-project-plugins-service.ts
```diff
@@ -567,6 +611,85 @@ export class NativeScriptProjectPluginsService implements IPluginsService {
...
+				packageJsonContent = this.getProjectPackageJsonContent().wait(),
+				pluginBasicInfo = this.getPluginBasicInformation(pluginIdentifier),
+				name = pluginBasicInfo.name,
+				selectedVersion = pluginBasicInfo.version || "latest",
+				basicPlugin = this.getBasicPluginInfoFromMarketplace(name, selectedVersion).wait() ||
+								this.getBasicPluginInfoFromNpm(name, selectedVersion).wait() ||
+								this.getBasicPluginInfoFromUrl(pluginIdentifier).wait();
+
+			if(!basicPlugin) {
+				this.$errors.failWithoutHelp(`Unable to add plugin ${pluginIdentifier}. Make sure you've provided a valid name, path to local directory or github URL.`);
```
teobugslayer: github -> GitHub

rosen-vladimirov: Maybe it should be `git` URL as we support urls like `https://github.com/Icenium/icenium-cli.git`
There's no requirement to be `github` only as far as I know. What do you think?

nadyaA:
baby-panda
:+1:

lib/services/nativescript-project-plugins-service.ts
teobugslayer: `git` is even better.

teobugslayer:
:+1:

PR #1336, opened 2016-01-19
Fix compilation of NS TypeScript projects

rosen-vladimirov:
When for NS TypeScript project user had called `npm install`, there are `*.d.ts` files inside `node_modules`.
But we also have definition files in `typings` directory, so the compilation fails.
Fix this by skipping `node_modules` dir when checking for `.ts` files. We do not want to compile files from `node_modules` dir.
Also add check if `tsconfig.json` exists to use it and just spawn tsc as all user options specified in tsconfig.json MUST be respected.

Fixes http://teampulse.telerik.com/view#item/306751

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -642,7 +642,11 @@ export class Project implements Project.IProject {
 
 	public getTypeScriptFiles(): IFuture<Project.ITypeScriptFiles> {
 		return ((): Project.ITypeScriptFiles => {
-			let projectFiles = this.$fs.enumerateFilesInDirectorySync(this.getProjectDir().wait());
+			let projectDir = this.getProjectDir().wait();
+			// Skip root's node_modules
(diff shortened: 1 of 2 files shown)


lib/project.ts
```diff
@@ -642,7 +642,8 @@ export class Project implements Project.IProject {
 
 	public getTypeScriptFiles(): IFuture<Project.ITypeScriptFiles> {
 		return ((): Project.ITypeScriptFiles => {
-			let projectFiles = this.$fs.enumerateFilesInDirectorySync(this.getProjectDir().wait());
+			let projectFiles = this.$fs.enumerateFilesInDirectorySync(this.getProjectDir().wait(),
```
teobugslayer: well, what about directory named `xxx-node_modules-xxx`?

teobugslayer:
:+1:

PR #1338, opened 2016-01-19
Fix plugins list for obosleted plugins

rosen-vladimirov:
When we have obsoleted plugin with the same identifier as its marketplace replacement, current code filters it.
For example in case you add `com.telerik.feedback` and execute `appbuilder plugin`, it will not be listed.
Filter only items where the replacement id is not the same as integrated plugin's id.


rosen-vladimirov:
ping @mbektchiev @Fatme

mbektchiev:
:+1:

PR #1344, opened 2016-01-25
Fix getting information for some nativescript plugins

rosen-vladimirov:
Fix getting information for nativescript plugins, which do not have `repository` property in `package.json`.
Try adding `plugin-var-plugin` and executing `appbuilder plugin` after that. Error is raised because the plugin does not have repository property and we are trying to get repository.url.

No regression potential.


Mitko-Kerezov:
:+1:

mbektchiev:
:+1:

endarova:
:+1:

PR #1345, opened 2016-01-25
Print information about plugin variables

rosen-vladimirov:
Print information about nativescript plugin variables.
Print the values when plugin is already added (`appbuilder plugin` command and when lisiting available plugins: `appbuilder plugin --available`).
Print plugins variables when listing available plugins (valid for Cordova and {N} Plugins) - even when plugin is not added, we know which are the required varialbes, show them to the user.

### DIFF
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -648,6 +648,30 @@ interface IPlugin {
 }
 
 /**
+ * Describes information about one plugin variable.
+ * In NativeScript project the variable is described in plugin's package.json with name and object as value.
+ * @example
+ * {
+ *    "name": "my-plugin",
+ *    "version": "1.0.0",
+ *    "nativescript": {
+ *        "variables": {
+ *           "my-var": {
+ *              "default": "defaultValue"
+ *           }
+ *        }
+ *    }
+ * }
+ */
+interface INativeScriptPluginVariable {
+	/**
+	 * The default value of the variable that will be used in case user does not specify value for the variable.
+	 * @optional
+	 */
+	defaultValue?: string;
+}
+
+/**
  * Describes basic information for each plugin.
  */
 interface IPluginInfoBase {
@@ -690,6 +714,11 @@ interface IPluginInfoBase {
 	 * Plugin specific identifier. It can differ from plugin's name.
 	 */
 	Identifier: string;
+
+	/**
+	 * Variables that have to be set in the application in order to be able to use the plugin correctly.
+	 */
+	Variables?: string[]|IDictionary<INativeScriptPluginVariable>;
 }
 
 interface IPluginVersion {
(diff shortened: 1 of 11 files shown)


lib/declarations.d.ts
```diff
@@ -648,6 +648,30 @@ interface IPlugin {
 }
 
 /**
+ * Describes information about one plugin variable.
+ * In NativeScript project the variable is describe in plugin's package.json with name and object as value.
```
teobugslayer: `is described`

lib/declarations.d.ts
```diff
@@ -690,6 +714,11 @@ interface IPluginInfoBase {
 	 * Plugin specific identifier. It can differ from plugin's name.
 	 */
 	Identifier: string;
+
+	/**
+	 * Variables, that have to be set in the application in order to be able to use the plugin correctly.
```
teobugslayer: the comma is not needed

lib/declarations.d.ts
```diff
@@ -690,6 +714,11 @@ interface IPluginInfoBase {
...
+	 */
+	Variables?: string[]|IDictionary<INativeScriptPluginVariable>;
```
teobugslayer: Why this is named with a capital letter? This means it is new'able.

lib/plugins-data.ts
```diff
@@ -55,24 +55,51 @@ export class CordovaPluginData implements IPlugin {
...
 					_.each(variableNames, (variableName:string) => {
 						output.push(util.format("        %s: %s", variableName, variables[variableName]));
 					});
 
 					result.push(output.join(EOL));
 				}
+			} else {
+				if(this.data.Variables) {
+					// We should never get here with anything that is not array, but anyway, lets assure we'll not throw some unexpected error.
+					if(_.isArray(this.data.Variables) && (<string[]>this.data.Variables).length) {
```
teobugslayer: I think it can be expressed more elegantly as
`if (this.data.Variables instanceof Array && this.data.Variables.length)` (see [type guards](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#type-guards)

lib/declarations.d.ts
rosen-vladimirov: not sure, but I've just moved it from one inteface to another. If we decide to change it, we'll have to manually check the code (just renaming does not work :cry: ). Do you want me to do it in this PR?

lib/plugins-data.ts
rosen-vladimirov: When trying to use this code:
```
Using tsc v1.7.5
d:/Work/icenium-cli/lib/plugins-data.ts(91,69): error TS2339: Property 'length' does not exist on type 'string[] | IDictionary<INativeScriptPluginVariable>'.
```

teobugslayer: Ah sure. Sorry for the noise.

teobugslayer:
:+1:

PR #1348, opened 2016-01-28
Fix ns plugin add

Mitko-Kerezov:
Write plugin identifier instead of name in package.json
In addition
* change/add messages upon plugin add/remove/fetch.
* allow installing plugins from `.tgz`
Ping @rosen-vladimirov

### DIFF
--- a/lib/services/nativescript-project-plugins-service.ts
+++ b/lib/services/nativescript-project-plugins-service.ts
@@ -110,18 +110,19 @@ export class NativeScriptProjectPluginsService implements IPluginsService {
 
 	public addPlugin(pluginIdentifier: string): IFuture<void> {
 		return (() => {
+			let pluginBasicInfo: IBasicPluginInformation;
 			if(this.isPluginInstalled(pluginIdentifier)) {
 				this.$logger.printMarkdown(util.format("Plugin `%s` is already installed.", pluginIdentifier));
 				return;
 			}
 
 			if(this.checkIsValidLocalPlugin(pluginIdentifier).wait()) {
-				this.installLocalPlugin(pluginIdentifier, {addPluginToPackageJson: true}).wait();
+				pluginBasicInfo = this.installLocalPlugin(pluginIdentifier, {addPluginToPackageJson: true}).wait();
 			} else {
-				this.setPluginInPackageJson(pluginIdentifier, {addPluginToPackageJson: true}).wait();
+				pluginBasicInfo = this.setPluginInPackageJson(pluginIdentifier, {addPluginToPackageJson: true}).wait();
 			}
 
-			this.$logger.printMarkdown(util.format("Successfully added plugin `%s`.", pluginIdentifier));
+			this.$logger.printMarkdown(util.format("Successfully added plugin `%s`.", pluginBasicInfo.name));
 		}).future<void>()();
 	}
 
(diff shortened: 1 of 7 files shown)


lib/services/nativescript-project-plugins-service.ts
```diff
@@ -110,18 +110,23 @@ export class NativeScriptProjectPluginsService implements IPluginsService {
...
 			if(this.isPluginInstalled(pluginIdentifier)) {
-				this.$logger.printMarkdown(util.format("Plugin `%s` is already installed.", pluginIdentifier));
+				this.$logger.printMarkdown(util.format("Plugin `%s` is already installed.", pluginIdentifier.toLowerCase()));
 				return;
 			}
 
-			if(this.checkIsValidLocalPlugin(pluginIdentifier).wait()) {
-				this.installLocalPlugin(pluginIdentifier, {addPluginToPackageJson: true}).wait();
+			if (path.extname(pluginIdentifier) === ".tgz") {
+				this.fetch(pluginIdentifier).wait();
```
rosen-vladimirov: hm, maybe I'm not correct, but `fetch` method installs the package to temp dir and copies it from there to plugins lib. After that it adds the plugin as local one from the plugins dir. So in the package.json you'll have:
```JSON
"dependencies": {
    "my-fetched-plugin": "file:plugins/myplugins"
}
```
Sooo this is not the expected behavior when I have `.tgz`. In fact we have to:
1) Try installing it to temp dir and check its package.json (get the id from package.json).
2) Check if the path to .tgz is in current project dir. If it is not, copy **the .tgz file** to the project dir
3) Add it to package.json using the file in project dir.

So we should receive something like:
```JSON
"dependencies": {
    "my-fetched-plugin": "file:plugins/myfile.tgz"
}
```

lib/services/nativescript-project-plugins-service.ts
```diff
@@ -179,6 +184,7 @@ export class NativeScriptProjectPluginsService implements IPluginsService {
 
 	public getPluginBasicInformation(pluginName: string): IBasicPluginInformation {
 		let [ name, version ] = pluginName.split("@");
+		name = name.toLowerCase();
```
rosen-vladimirov: what will happen in case I want to add this package:
https://www.npmjs.com/package/NodObjC
**NOT** this one: https://www.npmjs.com/package/nodobjc
For example `$ appbuilder plugin add NodeObjC@1.0.0`

lib/services/nativescript-project-plugins-service.ts
Mitko-Kerezov: I've decided against adding support for `.tgz` files in this PR - this will be handled in a separate one

Mitko-Kerezov:
Fixed PR as a whole, adding a TP Reference: [[CLI] Improper plugin name written in package,json](http://teampulse.telerik.com/view#item/309056)
Ping for a second review

Mitko-Kerezov:
run ci

rosen-vladimirov:
:+1: after green build

Mitko-Kerezov:
Unit tests were failing due to wrong project/item templates on the environment. 
Merging this PR with the [actual build's unit tests passing](http://bpc15:8080/job/ICE_CLI_PR/1505/console)

PR #1351, opened 2016-02-09
Appbuilder init discovers Ionic projects as hybrid ones

teobugslayer:
See http://teampulse.telerik.com/view#item/309474


rosen-vladimirov:
:+1:

PR #1352, opened 2016-02-15
Fix livesync --watch

rosen-vladimirov:
`appbuilder livesync --watch` fails with error when file is saved. The problem is in code used for NS CLI, which returns null instead of future.


teobugslayer:
:+1:

PR #1355, opened 2016-02-17
Fix overwriting DisplayName when iOSDisplayName is present and empty

Mitko-Kerezov:
If `iOSDisplayName` is present but an empty string, then our current logic sets the `DisplayName` to `iOSDisplayName` and right after that decides that no `DisplayName` is present and uses a default one.
Fix the issue, by checking if a `DisplayName` is already set and then fork the logic between `iOSDisplayName` and the default one.
Ping @teobugslayer @rosen-vladimirov

### DIFF
--- a/lib/project/framework-project-base.ts
+++ b/lib/project/framework-project-base.ts
@@ -64,6 +64,28 @@ export class FrameworkProjectBase implements Project.IFrameworkProjectBase {
 		return propertyValue;
 	}
 
+	public completeProjectProperties(properties: any): boolean {
+		let updated = false;
+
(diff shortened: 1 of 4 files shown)


lib/project/framework-project-base.ts
```diff
@@ -64,6 +64,30 @@ export class FrameworkProjectBase implements Project.IFrameworkProjectBase {
 		return propertyValue;
 	}
 
+
+
```
rosen-vladimirov: Multiple blank lines

rosen-vladimirov:
:+1:

PR #1356, opened 2016-02-17
Unify {N} samples' display names

Mitko-Kerezov:
Set display name to the actual name of the sample only.
TP Reference: [[CLI] Unify {N} samples' display names](http://teampulse.telerik.com/view#item/310120)
Ping @rosen-vladimirov @teobugslayer


teobugslayer:
:+1:

PR #1358, opened 2016-02-18
Set AndroidHardwareAcceleration to true

Mitko-Kerezov:
It is enabled by default in the new default Cordova version - 4.0.0
Ping @rosen-vladimirov


rosen-vladimirov:
:+1:

PR #1365, opened 2016-03-01
ProjectFilesProvider should extend ProjectFilesProviderBase

rosen-vladimirov:
After fix in common-lib, where a new ProjectFilesProviderBase has been added, ProjectFilesProvider now must extend ProjectFilesProviderBase.
Fix this, remove unused inteface in LiveSync command and remove `release` option as it's part of common options now.
Move common-options unit tests to common lib.


Mitko-Kerezov:
:+1:

PR #1366, opened 2016-03-01
Add tsconfig.json

rosen-vladimirov:
Add tsconfig.json so when CLI is opened with VS Code, the autocompletion will work!


Mitko-Kerezov:
:+1:

PR #1367, opened 2016-03-03
Extract more files when initializing project

rosen-vladimirov:
When `appbuilder init hybrid` is called, we should extract:
* `.abproject`
* `.debug.abproject`
* `.release.abproject`
* `.abignore`
* `cordova.ios.js`
* `cordova.android.js`
* `cordova.wp8.js`

When `appbuilder init native` is called, we should extract:
* `.abproject`
* `.abignore`
* `package.json`

In case any of them exists, we should not overwrite it.
Add new property to frameworkProject - `projectSpecificFiles` - it should define the required files that should exist after initialization.
Always extract all `*.abproject` and `*.abignore` files no matter of project type.
Add tests for the new functionality.
Introduce new project constant for `.abignore` file name.

Fixes http://teampulse.telerik.com/view#item/310769

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -243,22 +243,26 @@ export class Project implements Project.IProject {
 		return this.createFromTemplate(projectName, projectDir, template);
 	}
 
+	private get projectFilePatterns(): string[] {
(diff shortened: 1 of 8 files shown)


Mitko-Kerezov:
:+1:

PR #1369, opened 2016-03-11
Fix QR code generations for iOS Companion

rosen-vladimirov:
Remove unused property(servedApp) form iOSCompanion classes that breaks building for iOS companion apps.

http://teampulse.telerik.com/view#item/310864


Mitko-Kerezov:
:+1:

PR #1370, opened 2016-03-15
Import Ionic CLI projects

teobugslayer:
Only a spike

See http://teampulse.telerik.com/view#item/309474


rosen-vladimirov:
:+1:  after green build

PR #1377, opened 2016-03-24
Regenerate server-api

Mitko-Kerezov:
A minor bug inside our swagger did not allow for a clean api regeneration from endpoints containing multiple slashes (e.g. internal/apps/projects). Fix said bug and regenerate-api.
Api regeneration lead to some compilation issues - fix them by calling correct method overloads.
Ping @rosen-vladimirov


rosen-vladimirov:
:+1:

PR #1378, opened 2016-03-24
Fixed bug in commands fail if executed outside project.

TsvetanMilanov:
Created base ensure project command class to extract the common can execute method which checks if there is project when the command depends on project.
Refactored the other similar commands to remove the repeating can execute method.


TsvetanMilanov:
Close and retarget to release

PR #1379, opened 2016-03-24
Fixed the autocomplete

TsvetanMilanov:
When path to node has spaces (C:\Program Files) the child process exec cannot run the command. Need to place it in "" to run it without errors.


rosen-vladimirov:
:+1:  after green build

PR #1382, opened 2016-03-28
Move additional files in .ab folder

TsvetanMilanov:
When building the project with --download the configuration files for all platforms should be downloaded in .ab folder

Fixes http://teampulse.telerik.com/view#item/302390


lib/services/build.ts
```diff
@@ -1,4 +1,5 @@
 ///<reference path="../.d.ts"/>
+///<reference path="../common/.d.ts"/>
```
rosen-vladimirov: you do not need this

rosen-vladimirov:
:+1:

PR #1383, opened 2016-03-28
Fix plugin configure command

TsvetanMilanov:
When the .debug.abproject and the .release.abproject files are deleted and then plugin is installed without specified configuration and the plugin variables are set in the .abproject file. When configuring the plugin variables for specific configuration the original plugin variables must be cloned because the changes must be saved in the configuration specific file only.
When getting the plugin variable from var option the configuration needs to be checked because if there is no specified configuration in the command it will be undefined.

Fixes http://teampulse.telerik.com/view#item/282804

### DIFF
--- a/lib/services/cordova-project-plugins-service.ts
+++ b/lib/services/cordova-project-plugins-service.ts
@@ -375,7 +375,8 @@ export class CordovaProjectPluginsService implements IPluginsService {
 		return (() => {
 			let plugin = this.getBestMatchingPlugin(pluginName, version);
 			let pluginData = <IMarketplacePluginData>plugin.data;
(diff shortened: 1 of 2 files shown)


lib/services/cordova-project-plugins-service.ts
```diff
@@ -694,7 +694,7 @@ export class CordovaProjectPluginsService implements IPluginsService {
 	 */
 	private getPluginVariableFromVarOption(variableName: string, configuration: string): any {
 		let varOption = this.$options.var;
-		configuration = configuration.toLowerCase();
+		configuration = configuration ? configuration.toLowerCase() : undefined;
```
rosen-vladimirov:
```TypeScript
configuration = configuration && configuration.toLowerCase();
```

rosen-vladimirov:
:+1:

PR #1384, opened 2016-04-01
Fix framework not supported error message

TsvetanMilanov:
Add supported frameworks versions to the error message when using appbuilder mobileframework and the current version is not supported.

Fixes http://teampulse.telerik.com/view#item/312441


rosen-vladimirov:
:+1:

PR #1385, opened 2016-04-04
 Workaround error 21 when working with iOS devices

rosen-vladimirov:
When CLI is trying to work with different devices, calling setInterval to detect new devices, while deploying/livesyncing on currently attached iOS device is throwing error 21 on MacOS.
For example: `appbuilder deploy` when only iOS device is attached.
Workaround this by detecting the platform first and call setInterval only in very rare cases.


rosen-vladimirov:
:+1:  from @TsvetanMilanov

PR #1387, opened 2016-04-06
Fix livesync issues when app is not installed on device

rosen-vladimirov:
* Fix LiveSync issue for NativeScript iOS projects, when application is not installed on device and `$ appbuilder livesync ios` is called - this should install the application. Instead it does nothing.
* Fix LiveSync to companion app when application is not installed on device. Currently it was trying to install the application instead of sending the app to companion.


TsvetanMilanov:
:+1:

PR #1388, opened 2016-04-06
Fix bug in commands fail if executed outside project

TsvetanMilanov:
When checking if command is executed in project in the typescript-compilation hook need to return not to throw error because the not in project message and command help will not be shown.

Fixes http://teampulse.telerik.com/view#item/308947


TsvetanMilanov:
run ci

rosen-vladimirov:
:+1:

rosen-vladimirov:
ping @tytodorov

tytodorov:
:+1:

PR #1391, opened 2016-04-08
Fix LiveSync to Companion apps

rosen-vladimirov:
Use correct LiveSync urls for Companion apps.
Pass App Identifier and configuration.

Fixes http://teampulse.telerik.com/view#item/313027


TsvetanMilanov:
:+1:

tytodorov:
AppId should be project name, not ApplicationIdentifier

rosen-vladimirov:
@tytodorov : Fixed, thanks for pointing this :)

tytodorov:
:+1: After green build

PR #1400, opened 2016-04-28
Change plugman's registry to use npm registry

TsvetanMilanov:
We need to use npm registry when searching for cordova plugins because cordova registry is unavailable.


lib/services/cordova-plugins.ts
```diff
@@ -103,7 +104,9 @@ export class CordovaPluginsService implements ICordovaPluginsService {
 
 	public configure(): void {
 		let future = new Future();
-		let params = ["set", "registry", this.$config.CORDOVA_PLUGINS_REGISTRY];
+
+		// Need to set plugman's registry to use npm registry to avoid problems with cordova registry availability.
+		let params = ["set", "registry", "http://registry.npmjs.org"];
```
rosen-vladimirov: Can't we just change the value of the variable, I believe it's written in config-base.json

rosen-vladimirov:
:+1:  after green build

PR #1401, opened 2016-05-03
Fix remote command

rosen-vladimirov:
Due to change in the names that AppBuilder CLI passes to ios-sim-portable, the remote command is not working when user's XCode version is 7.x
The current names work for XCode 6.x only.
Check the version of the XCode on the machine and pass correct names.


TsvetanMilanov:
run ci

TsvetanMilanov:
:+1:

PR #1403, opened 2016-05-05
Clean unused code for already deleted {N} versions

TsvetanMilanov:
Since NativeScript version 1.5.2 is obsolete and it cannot be migrated to the old logic for migrating NativeScript projects is not needed any more.
Also the server returns NativeScript.json with additional property - obsoleteVersions and the property npmVersions is renamed to versions.

The improvement is described here http://teampulse.telerik.com/view#item/312022

### DIFF
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -450,9 +450,6 @@ interface INativeScriptMigrationConfiguration {
 	appResourcesRequiredPath: string;
 	appResourcesObsoletePath: string;
 
-	valuesStylesXmlPath: string;
-	valuesV21StylesXmlPath: string;
-
 	shouldRollBackAppResources: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


TsvetanMilanov:
run ci

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "appbuilder",
   "preferGlobal": true,
-  "version": "3.2.0",
```
rosen-vladimirov: why is this changed?

rosen-vladimirov:
we should delete the code related to manual migration of `styles.xml`. Can you please delete it as well?

rosen-vladimirov:
Cannot be merged without  https://github.com/Icenium/Ice/pull/5590

rosen-vladimirov:
run ci

rosen-vladimirov:
:+1:

TsvetanMilanov:
run ci

PR #1405, opened 2016-05-12
Fix post install

rosen-vladimirov:
Currently post install fails as we are trying to call `.toLowerCase()` of undefined.


Mitko-Kerezov:
Seems legit :+1:

PR #1408, opened 2016-05-16
Allow starting of NativeScript apps on Android via Proton

rosen-vladimirov:
Currently when deploying NativeScript apps on Android via Proton, we are unable to start the application, as we pass incorrect StartPackageActivity.
Fix this by extracting the correct names in constants. When deploy is called from Proton side, we'll set the correct name in the `$project`.
After that the execution of start method will use the value from `$staticConfig`. StaticConfig will return the value of project's property.

`deployOnDevices` method now has additional argument - framework. If it is passed and the provided value is NativeScript, we'll use the NativeScript's StartPackageActivity.
In all other cases we'll use Cordova's StartPackageActivity.

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -1,7 +1,7 @@
 ///<reference path=".d.ts"/>
 "use strict";
 import * as path from "path";
-import {StaticConfigBase} from "./common/static-config-base";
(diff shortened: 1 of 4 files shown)


Mitko-Kerezov:
:+1:

TsvetanMilanov:
:+1:

PR #1415, opened 2016-05-25
Fix QR Code generation for projects with long names

TsvetanMilanov:
When we try to generate QR Code for project with long name the library we use fails.
The QR Code generation logic was not compatible with the library and the exception that was thrown was not expected.


rosen-vladimirov:
:+1:

PR #1417, opened 2016-05-25
Fix Ionic project backup

TsvetanMilanov:
When creating the backup of the project the AppBuilder project files are already added to the original project directory and they need to be excluded from the backup.

### DIFF
--- a/lib/ionic-project-transformator.ts
+++ b/lib/ionic-project-transformator.ts
@@ -4,6 +4,7 @@
 import * as path from "path";
 import * as shelljs from "shelljs";
 import * as xmlMapping from "xml-mapping";
+import {EOL} from "os";
 
 export class IonicProjectTransformator implements IIonicProjectTransformator {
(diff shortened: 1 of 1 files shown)


lib/ionic-project-transformator.ts
```diff
@@ -416,6 +438,28 @@ export class IonicProjectTransformator implements IIonicProjectTransformator {
 		});
 		return folderNames;
 	}
+
+	private isAppBuilderProjectFile(file: string): boolean {
+		let isAppBuilderFile = false;
+		_.each(this.appbuilderProjectFiles, (appBuilderFileName: string) => {
+			if (file.indexOf(appBuilderFileName) >= 0) {
+				isAppBuilderFile = true;
+				return;
```
rosen-vladimirov:
```TypeScript
return false;
```
To break the each :)

rosen-vladimirov:
:+1:

PR #1418, opened 2016-05-26
Change the properties for NativeScriptMigrationData

TsvetanMilanov:
The npmVersions property is removed, obsoleteVersions is renamed to deprecatedVersions.


rosen-vladimirov:
:+1:

PR #1419, opened 2016-05-26
Update message for Node.js 0.10.x deprecation

rosen-vladimirov:
Update message for Node.js 0.10.x deprecation and add additional arguments to verification method - CLI's name and version in which support for Node.js 0.10.x will be removed.

Remove 0.10.x versions from valid versions in package.json.


TsvetanMilanov:
:+1:

PR #1422, opened 2016-05-27
Add EOL in .abignore file in Ionic projects

TsvetanMilanov:
When creating backup of the Ionic project and adding the Ionic_Backup folder to the .abignore file need to add EOL at the and of it.


rosen-vladimirov:
:+1:

PR #1423, opened 2016-05-27
Kill hanging processes of ios-webkit-debug-proxy

TsvetanMilanov:
Sometimes the ios-webkit-debug-proxy is leaving hanging processes which need to be killed when the user stops debugging (press Ctrl + C).


TsvetanMilanov:
Closing because we decided not to use ios-webkit-debug-proxy for debugging iOS cordova applications on Mac.

PR #1424, opened 2016-05-30
Fix Ionic config.xml resources cloning

TsvetanMilanov:
Need to check if the Ionic config.xml for each registered platform has registered splash screen and if not we should not try to copy it.


rosen-vladimirov:
:+1:

TsvetanMilanov:
run ci

PR #1428, opened 2016-05-31
Fix tracking of users

rosen-vladimirov:
We are using incorrect property - id, now it's called uid. This leads to tracking all users with the same user id - "undefined".
Introduce interfaces for User and Tenant information.


Mitko-Kerezov:
👍 but we have to remember to remove this commit when the server's swagger is upgraded

PR #1429, opened 2016-05-31
Check if livesync is supported

TsvetanMilanov:
When checking if the livesync is supported and the application is not installed the result is false and we need to deploy the application on the device and check again.


TsvetanMilanov:
Closing and retarget to master.

PR #1433, opened 2016-06-01
Delete files when LiveSync on Android

rosen-vladimirov:
When `livesync android --watch` is used and user deletes a file, we do not actually delete it. Fix this by sending correct command to the LiveSync plugin.
Add public method in liveSyncService that should be used to delete files when LiveSync operation is in progress.
This way, when user deletes file in Proton, the application will call `liveSyncService.deleteFiles` method for it and it will be removed from device.


TsvetanMilanov:
:+1:

PR #1434, opened 2016-06-02
Replace shelljs cp with fs method

TsvetanMilanov:
For some reason `shelljs.cp` is really slow and that's why we should use the `copyDir` method in our fs.


nirinchev:
Do we need the console.logs and the commented-out old implementation?

TsvetanMilanov:
run ci

TsvetanMilanov:
Closed because the fs method is not faster than the shelljs cp.

PR #1435, opened 2016-06-02
Increase tests timeout

TsvetanMilanov:
Need to increase the tests timeout because in some cases the tests for the Ionic project transformation are really slow.


rosen-vladimirov:
:+1:  after green build

rosen-vladimirov:
run ci

PR #1437, opened 2016-06-03
Fix device run command

TsvetanMilanov:
When executing `appbuilder device run appid` outside of project the CLI cannot determine which startPackageActivity name to use - for NativeScript or for Cordova application. The fix is to check where the command is executed and if it is not in project we need to run the adb command for both NativeScript and Cordova projects.


TsvetanMilanov:
run ci

TsvetanMilanov:
run ci

TsvetanMilanov:
Close and retarget to master.

PR #1439, opened 2016-06-03
Fix iOS status bar style message for Hidden value

TsvetanMilanov:
The Hidden value for iOS status bar style is supported only on iOS versions lower than iOS 7.

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -20,6 +20,9 @@ export class Project implements Project.IProject {
 		"javascript.blank": "Blank"
 	};
 
+	// The json schema on the server has wrong description for the Hidden property of the iOS Status Bar Style.
+	private static HIDDEN_IOS_STATUS_BAR_DESCRIPTION = "Hidden";
+
 	private _hasBuildConfigurations: boolean = false;
 	private _projectSchema: any;
 	private cachedProjectDir: string = "";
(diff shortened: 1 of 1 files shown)


lib/project.ts
```diff
@@ -19,6 +19,7 @@ export class Project implements Project.IProject {
 		"kendoui.blank": "KendoUI.Empty",
 		"javascript.blank": "Blank"
 	};
+	private static WRONG_IOS_STATUS_BAR_DESCRIPTION = "Hidden";
```
nirinchev: The name of this field seems off - there's nothing wrong or descriptive about the word "Hidden" 😺

rosen-vladimirov: I Agree, `HIDDEN_IOS_STATUS_BAR_DESCRIPTION` seems better

rosen-vladimirov:
So, if I use `appbuilder prop set iOSStatusBar "Hidden (iOS 6.1 and earlier)"`, as I've seen it from `$ appbuilder prop print iOSStatusBar --validValues`, will it work for me ?

TsvetanMilanov:
Closing because Cordova 4.0 has minimum deployment target iOS 8.0 and this bug is for versions earlier than iOS 6.1,  it does not need to be fixed.

PR #1444, opened 2016-06-08
Improve TypeScript files

rosen-vladimirov:
Improve .ts files by removing the references and "use strict" as it is added by default from TypeScript compiler.

Update TypeScript and linter


TsvetanMilanov:
:+1:

PR #1446, opened 2016-06-10
Implement configuration logic

Mitko-Kerezov:
Includes changes for
* Prop set/print commands
* Build
* LiveSync (both regular and companion)

TP Reference: [Support environment configurations](http://teampulse.telerik.com/view#item/308175)
Ping @rosen-vladimirov @TsvetanMilanov for a **thorough** review

Tests are on the way!

### DIFF
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -1,49 +1,56 @@
-import commonOptionsLibPath = require("./common/options");
+import {OptionType, OptionsBase} from "./common/options";
+import {Configurations} from "./common/constants";
 import * as osenv from "osenv";
 import * as path  from "path";
 
-let OptionType = commonOptionsLibPath.OptionType;
-
-export class Options extends commonOptionsLibPath.OptionsBase {
+export class Options extends OptionsBase {
 	constructor($errors: IErrors,
 		$staticConfig: IStaticConfig,
 		$hostInfo: IHostInfo) {
 		super({
-			all: { type: OptionType.Boolean},
-			answers: { type: OptionType.String },
-			available: { type: OptionType.Boolean },
-			certificate: { type: OptionType.String  },
-			companion: { type: OptionType.Boolean },
-			core: { type: OptionType.Boolean },
-			count: { type: OptionType.Number },
-			debug: { type: OptionType.Boolean, alias: "d" },
-			deploy: { type: OptionType.String },
-			download: { type: OptionType.Boolean },
-			force: { type: OptionType.Boolean, alias: "f" },
-			group: { type: OptionType.Array },
-			icon: { type: OptionType.String},
-			latest: { type: OptionType.Boolean },
-			mandatory: { type: OptionType.Boolean},
-			professional: { type: OptionType.Boolean },
-			provision: { type: OptionType.String  },
-			public: { type: OptionType.Boolean },
-			publish: { type: OptionType.Boolean },
-			saveTo: { type: OptionType.String},
-			screenBuilderCacheDir: { type: OptionType.String },
-			sendEmail: { type: OptionType.Boolean },
-			sendPush: { type: OptionType.Boolean },
-			simulator: { type: OptionType.Boolean, default: true},
-			skipUi: { type: OptionType.Boolean },
-			splash: { type: OptionType.String},
-			template: { type: OptionType.String, alias: "t" },
-			validValue: { type: OptionType.Boolean },
-			verified: { type: OptionType.Boolean }
-		},
-		path.join($hostInfo.isWindows ? process.env.LocalAppData : path.join(osenv.home(), ".local/share"), "Telerik", "BlackDragon", ".appbuilder-cli"),
+				all: { type: OptionType.Boolean},
+				answers: { type: OptionType.String },
+				available: { type: OptionType.Boolean },
+				certificate: { type: OptionType.String  },
+				companion: { type: OptionType.Boolean },
+				core: { type: OptionType.Boolean },
+				count: { type: OptionType.Number },
+				deploy: { type: OptionType.String },
+				download: { type: OptionType.Boolean },
+				force: { type: OptionType.Boolean, alias: "f" },
+				group: { type: OptionType.Array },
+				icon: { type: OptionType.String},
+				latest: { type: OptionType.Boolean },
+				mandatory: { type: OptionType.Boolean},
+				professional: { type: OptionType.Boolean },
+				provision: { type: OptionType.String  },
+				public: { type: OptionType.Boolean },
+				publish: { type: OptionType.Boolean },
+				saveTo: { type: OptionType.String},
+				screenBuilderCacheDir: { type: OptionType.String },
+				sendEmail: { type: OptionType.Boolean },
+				sendPush: { type: OptionType.Boolean },
+				simulator: { type: OptionType.Boolean, default: true},
+				skipUi: { type: OptionType.Boolean },
+				splash: { type: OptionType.String},
+				template: { type: OptionType.String, alias: "t" },
+				validValue: { type: OptionType.Boolean },
+				verified: { type: OptionType.Boolean }
+			},
+			path.join($hostInfo.isWindows ? process.env.LocalAppData : path.join(osenv.home(), ".local/share"), "Telerik", "BlackDragon", ".appbuilder-cli"),
 			$errors, $staticConfig);
 
 		let that = <any>this;
 		that.screenBuilderCacheDir = path.join((($hostInfo.isWindows && this.defaultProfileDir === that.profileDir) ? path.join(process.env.LocalAppData, "Telerik"): that.profileDir), "sb");
+		if (that.release) {
+			that.config = that.config || [];
+			that.config.push(Configurations.Release);
+		}
+
+		if (that.debug) {
+			that.config = that.config || [];
+			that.config.push(Configurations.Debug);
+		}
 	}
 }
 $injector.register("options", Options);
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -2,54 +2,50 @@ import {EOL} from "os";
 import * as path from "path";
 import * as util from "util";
 import * as commonHelpers from "./common/helpers";
+import {Configurations} from "./common/constants";
+import {ProjectBase} from "./common/appbuilder/project/project-base";
 import Future = require("fibers/future");
 import * as helpers from "./helpers";
-import { TARGET_FRAMEWORK_IDENTIFIERS } from "./common/mobile/constants";
+import { TARGET_FRAMEWORK_IDENTIFIERS } from "./common/constants";
 
-export class Project implements Project.IProject {
-	private static JSON_PROJECT_FILE_NAME_REGEX = "[.]abproject";
+export class Project extends ProjectBase implements Project.IProject {
 	private static CHUNK_UPLOAD_MIN_FILE_SIZE = 1024 * 1024 * 50;
-	private static CONFIGURATION_FILE_SEARCH_PATTERN: RegExp = new RegExp(".*.abproject$", "i");
-	private static VALID_CONFIGURATION_CHARACTERS_REGEX = "[-_A-Za-z0-9]";
-	private static CONFIGURATION_FROM_FILE_NAME_REGEX = new RegExp("^[.](" + Project.VALID_CONFIGURATION_CHARACTERS_REGEX + "+?)" + Project.JSON_PROJECT_FILE_NAME_REGEX + "$", "i");
 	private static INDENTATION = "     ";
 	private static UI_TEMPLATE_NAMES: IStringDictionary = {
 		"kendoui.blank": "KendoUI.Empty",
 		"javascript.blank": "Blank"
 	};
 
-	private _hasBuildConfigurations: boolean = false;
 	private _projectSchema: any;
 	private cachedProjectDir: string = "";
 
 	private frameworkProject: Project.IFrameworkProject;
-	public projectData: Project.IData;
-	public configurationSpecificData: IDictionary<Project.IData>;
 	public get projectDir(): string {
 		return this.getProjectDir().wait();
 	}
 
 	constructor(private $config: IConfiguration,
-		private $errors: IErrors,
 		private $frameworkProjectResolver: Project.IFrameworkProjectResolver,
-		private $fs: IFileSystem,
+		private $ionicProjectTransformator: IIonicProjectTransformator,
 		private $jsonSchemaValidator: IJsonSchemaValidator,
 		private $loginManager: ILoginManager,
-		private $logger: ILogger,
 		private $multipartUploadService: IMultipartUploadService,
 		private $progressIndicator: IProgressIndicator,
-		private $projectConstants: Project.IConstants,
 		private $projectFilesManager: IProjectFilesManager,
 		private $projectPropertiesService: IProjectPropertiesService,
 		private $server: Server.IServer,
-		private $staticConfig: IStaticConfig,
 		private $templatesService: ITemplatesService,
 		private $prompter: IPrompter,
-		private $options: IOptions,
-		private $ionicProjectTransformator: IIonicProjectTransformator) {
-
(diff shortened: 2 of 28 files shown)


lib/options.ts
```diff
@@ -1,49 +1,56 @@
...
+				template: { type: OptionType.String, alias: "t" },
+				validValue: { type: OptionType.Boolean },
+				verified: { type: OptionType.Boolean }
+			},
+			path.join($hostInfo.isWindows ? process.env.LocalAppData : path.join(osenv.home(), ".local/share"), "Telerik", "BlackDragon", ".appbuilder-cli"),
 			$errors, $staticConfig);
 
 		let that = <any>this;
 		that.screenBuilderCacheDir = path.join((($hostInfo.isWindows && this.defaultProfileDir === that.profileDir) ? path.join(process.env.LocalAppData, "Telerik"): that.profileDir), "sb");
+		if (that.release) {
```
TsvetanMilanov: Can we use `this` instead of `that` here?

lib/project.ts
```diff
@@ -392,12 +388,17 @@ export class Project implements Project.IProject {
...
+					});
+				} else {
 				this.$projectPropertiesService.updateProjectProperty(this.projectData, undefined, mode, normalizedPropertyName, propertyValues).wait();
-				_.each(this.configurationSpecificData, configSpecificData => this.$projectPropertiesService.removeProjectProperty(configSpecificData, normalizedPropertyName, this.projectData));
+			}
 			}
 
 			this.saveProject(this.getProjectDir().wait(), projectConfigurations).wait();
-			this.printProjectProperty(normalizedPropertyName).wait();
+			this.printProjectProperty(normalizedPropertyName, projectConfigurations[0]).wait();
```
TsvetanMilanov: I think it will be better if in the message for the updated property we include all the project configurations that are updated because if the user wants to change the Author property for multiple project configurations and the message after the command execution is `The value of Author is: SomeAuthor` it can be confusing - is it changed only in .abproject or in all project configuration files or is it changed for the specified configurations.

lib/project.ts
```diff
@@ -368,22 +373,13 @@ export class Project implements Project.IProject {
...
 		}
 
-		return configurations;
-	}
-
 	public updateProjectPropertyAndSave(mode: string, propertyName: string, propertyValues: string[]): IFuture<void> {
 		return (() => {
 			this.ensureProject();
-			let projectConfigurations = _.keys(this.configurationSpecificData);
+			let projectConfigurations = this.getConfigurationsSpecifiedByUser();
```
TsvetanMilanov: If the user enters invalid configuration we should display warning which says that the invalid configuration does not exist and the changes for it will not be applied or throw an error with message for invalid configuration. Now if an invalid configuration is passed the command will continue and print successful message but there will be no result from the command.

lib/services/build.ts
```diff
@@ -97,10 +97,11 @@ export class BuildService implements Project.IBuildService {
 	private requestCloudBuild(settings: Project.IBuildSettings): IFuture<Project.IBuildResult> {
 		return ((): Project.IBuildResult => {
 			settings.platform = this.$mobileHelper.normalizePlatformName(settings.platform);
-			let projectData = this.$project.projectData;
+			let projectData = this.$project.projectInformation.configurationSpecificData[settings.projectConfiguration.toLowerCase()];
```
TsvetanMilanov: If we want to support case sensitive project configurations we do not need this `.toLowerCase()`

lib/services/livesync/livesync-service.ts
```diff
@@ -34,16 +37,61 @@ export class LiveSyncService implements ILiveSyncService {
...
+				livesyncData.configuration = deviceConfigurationInfos && deviceConfigurationInfos[0] && deviceConfigurationInfos[0].configuration;
+				if (livesyncData.configuration) {
+					this.$options.config = [livesyncData.configuration];
+				}
+
 			$liveSyncServiceBase.sync(livesyncData).wait();
+			} else {
+				configurations.forEach(configuration => {
+					livesyncData.configuration = configuration;
+					livesyncData.appIdentifier = this.$project.projectInformation.configurationSpecificData[configuration.toLowerCase()].AppIdentifier;
```
TsvetanMilanov: If we want to support case sensitive project configurations we do not need this `.toLowerCase()`

lib/services/build.ts
```diff
@@ -125,7 +126,7 @@ export class BuildService implements Project.IBuildService {
 				} else if (settings.buildForTAM) {
 					this.$logger.warn("You have not specified certificate to code sign this app. We'll use default debug certificate. " +
 						"Use --certificate option to specify your own certificate. You can check available certificates with '$ appbuilder certificate' command.");
-				} else if (settings.configuration === "Release") {
+				} else if (settings.buildConfiguration === "Release" ) {
```
TsvetanMilanov: Can we use `constants.Configurations.Release` instead of `"Release"` here?

lib/project.ts
```diff
@@ -368,22 +373,13 @@ export class Project implements Project.IProject {
...
 
 	public getConfigurationsSpecifiedByUser(): string[] {
-		let configurations: string[] = [];
-		if (this.$options.debug) {
-			configurations.push(this.$projectConstants.DEBUG_CONFIGURATION_NAME);
-		}
-
-		if (this.$options.release) {
-			configurations.push(this.$projectConstants.RELEASE_CONFIGURATION_NAME);
+		return this.$options.config || [];
```
TsvetanMilanov: If the user enters invalid configuration we should display warning which says that the invalid configuration does not exist and the changes for it will not be applied or throw an error with message for invalid configuration. If we display warning we should remove the invalid configurations and when this method is used it will return only valid configurations.

lib/services/build.ts
```diff
@@ -317,8 +318,11 @@ export class BuildService implements Project.IBuildService {
 			this.$jsonSchemaValidator.validate(this.$project.projectData);
 			this.$jsonSchemaValidator.validateWithBuildSchema(this.$project.projectData, settings.platform);
 
-			settings.configuration = settings.configuration || (this.$options.release ? "Release" : "Debug");
-			this.$logger.info("Building project for platform '%s', configuration '%s'", settings.platform, settings.configuration);
+			settings.projectConfiguration = settings.projectConfiguration || (this.$options.config && this.$options.config[0]) || constants.Configurations.Debug;
```
TsvetanMilanov: If `this.$options.config[0]` is invalid project configuration the build will fail with `Cannot read property 'AppIdentifier' of undefined`. Check if the configuration is valid and if it is not valid - throw an error with suitable message.

TsvetanMilanov:
The livesync for companion app does not work when executing `appbuilder livesync --config test --companion` but it works when executing `appbuilder livesync --companion --debug`

lib/project.ts
```diff
@@ -70,8 +66,22 @@ export class Project implements Project.IProject {
 		return this.frameworkProject.configFiles;
 	}
 
-	public get startPackageActivity(): string {
-		return this.frameworkProject.startPackageActivity;
+	public get projectData(): Project.IData {
+		if (!this._projectData) {
+			this.readProjectData().wait();
+	}
```
rosen-vladimirov: this `}` is not on the correct place.

lib/project.ts
```diff
@@ -629,7 +630,9 @@ export class Project implements Project.IProject {
 		return (() => {
 			let configs = (configurations && configurations.length > 0) ? configurations : this.configurations;
 			projectDir = projectDir || this.getProjectDir().wait();
+			if (!configurations || !configurations.length) {
 			this.$fs.writeJson(path.join(projectDir, this.$staticConfig.PROJECT_FILE_NAME), this.projectData).wait();
```
rosen-vladimirov: tab to the right maybe

lib/services/livesync/livesync-service.ts
```diff
@@ -34,16 +37,61 @@ export class LiveSyncService implements ILiveSyncService {
...
+				if (deviceConfigurations && _.uniq(deviceConfigurationInfos, dci => dci.configuration).length !== 1) {
+						this.$errors.failWithoutHelp("Cannot LiveSync because application is deployed with different configurations across the devices.", deviceConfigurations);
+				}
+
+				livesyncData.configuration = deviceConfigurationInfos && deviceConfigurationInfos[0] && deviceConfigurationInfos[0].configuration;
+				if (livesyncData.configuration) {
+					this.$options.config = [livesyncData.configuration];
+				}
+
 			$liveSyncServiceBase.sync(livesyncData).wait();
```
rosen-vladimirov: tab to the right

lib/services/build.ts
rosen-vladimirov: and shouldn't we `toLowerCase()` everything

lib/project.ts
```diff
@@ -707,100 +710,26 @@ export class Project implements Project.IProject {
...
-						if (_.has(data, "projectType")) {
-							data["Framework"] = data["projectType"];
-							delete data["projectType"];
-						} else {
-							data["Framework"] = TARGET_FRAMEWORK_IDENTIFIERS.Cordova;
+	private getProjectRelativePath(fullPath: string, projectDir: string): string {
+		projectDir = path.join(projectDir, path.sep);
+		if (!_.startsWith(fullPath, projectDir)) {
+			throw new Error("File is not part of the project.");
 						}
```
rosen-vladimirov: tab to the left maybe

lib/project.ts
```diff
@@ -392,12 +388,17 @@ export class Project implements Project.IProject {
 			if (normalizedPropertyName === this.$projectConstants.CORE_PLUGINS_PROPERTY_NAME) {
 				this.$projectPropertiesService.updateCorePlugins(this.projectData, this.configurationSpecificData, mode, propertyValues, this.getConfigurationsSpecifiedByUser()).wait();
 			} else {
+				if (projectConfigurations.length) {
+					_.each(projectConfigurations, configuration => {
+						this.$projectPropertiesService.updateProjectProperty(this.projectData, this.configurationSpecificData[configuration], mode, normalizedPropertyName, propertyValues).wait();
+					});
+				} else {
 				this.$projectPropertiesService.updateProjectProperty(this.projectData, undefined, mode, normalizedPropertyName, propertyValues).wait();
```
rosen-vladimirov: Indent me :crying_cat_face:

lib/project.ts
```diff
@@ -707,100 +710,26 @@ export class Project implements Project.IProject {
...
-							this._hasBuildConfigurations = true;
-						}
-					});
-				} catch (err) {
-					if (err === "FUTURE_PROJECT_VER") {
-						this.$errors.fail({
-							formatStr: "This project is created by a newer version of AppBuilder. Upgrade AppBuilder CLI to work with it.",
-							suppressCommandHelp: true
-						});
+		return fullPath.substring(projectDir.length);
```
rosen-vladimirov: is there any chance fullPath to be undefined or null ?

lib/services/livesync/livesync-service.ts
```diff
@@ -34,16 +37,61 @@ export class LiveSyncService implements ILiveSyncService {
...
+						let configInfo = device.getApplicationInfo(livesyncData.appIdentifier).wait();
+						if (configInfo) {
+							deviceConfigurationInfos.push(configInfo);
+						}
+					}).future<void>()()).wait();
 
+					livesyncData.canExecute = (device: Mobile.IDevice): boolean => {
+						let deviceConfigurationInfo = _.find(deviceConfigurationInfos, dci => dci.deviceIdentifier === device.deviceInfo.identifier);
+						if (deviceConfigurationInfo && deviceConfigurationInfo.configuration && deviceConfigurationInfo.configuration.toLowerCase() !== configuration.toLowerCase() && !this.$options.companion) {
+							this.$logger.warn(`LiveSync will not be performed on device with identifier ${device.deviceInfo.identifier}. You are trying to LiveSync in configuration ${configuration} but the device was deployed in configuration ${deviceConfigurationInfo.configuration}.`);
```
rosen-vladimirov: This message does not seem correct. The device itself is not deployed.
Can we remove the **in** -> You are trying to LiveSync **in** configuration
Maybe just:
```TypeScript
You are trying to LiveSync configuration ${configuration}, but the application is deployed with ${deviceConfigurationInfo.configuration} configuration.
```

lib/options.ts
Mitko-Kerezov: Apparently not - `this` is of type `Options` and we dynamically attach properties to `Options`. In other words - `this` doesn't have any properties so the code won't compile

lib/services/build.ts
Mitko-Kerezov: I think we'd rather we didn't support case sensitive configurations - I'll fix the code around so that we explicitly support case-insensitive configurations

lib/services/build.ts
Mitko-Kerezov: @TsvetanMilanov valid point

@rosen-vladimirov 
`settings.buildConfiguration` is either `constants.Configurations.Release` or `constants.Configurations.Debug` so no lowercase is required - I can add it if you insist, however it's unnecessary

lib/project.ts
Mitko-Kerezov: don't think so, but even if that were the case, then an error would be thrown a little above, because
`_.startsWith(fullPath, projectDir)` would return `false`

lib/services/livesync/livesync-service.ts
nirinchev: Or even better, ping @ikoevska 😸

ikoevska: `Cannot LiveSync to device with identifier ${device.deviceInfo.identifier}. You are trying to synchronize changes for the ${configuration} configuration but the device expects changes for the ${deviceConfigurationInfo.configuration} configuration. Change the target configuration for the LiveSync operation or re-build and re-deploy your app in another configuration.`

If this is way too long, consider removing the suggested resolution. However, I believe that error messages should suggest a resolution instead of just stating a fact.

Mitko-Kerezov: @ikoevska 
I'll stick with the long message - it is a bit long, however I like the idea of a solution within the error message - thanks

Mitko-Kerezov:
@TsvetanMilanov 
`appbuilder livesync --config test --companion` seems to be working - not sure whether PR comment fixes have something to do with the solution though - can you confirm it works?

lib/services/livesync/livesync-service.ts
```diff
@@ -34,16 +37,61 @@ export class LiveSyncService implements ILiveSyncService {
...
+				excludedProjectDirsAndFiles: this.excludedProjectDirsAndFiles,
+				additionalConfigurations: this.$project.projectInformation.configurations
 			};
 
-			$liveSyncServiceBase.sync(livesyncData).wait();
+			let deviceConfigurationInfos: Mobile.IApplicationInfo[] = [];
 
+			let configurations = this.$project.getConfigurationsSpecifiedByUser();
+			if (!configurations.length) {
+				this.$devicesService.execute(device => (() => {
```
rosen-vladimirov: Please extract this to separate method as you are using it in the "else" as well.

rosen-vladimirov: and you can call it before `if-else`

lib/project.ts
```diff
@@ -368,22 +373,27 @@ export class Project implements Project.IProject {
 	}
 
 	public getConfigurationsSpecifiedByUser(): string[] {
-		let configurations: string[] = [];
-		if (this.$options.debug) {
-			configurations.push(this.$projectConstants.DEBUG_CONFIGURATION_NAME);
+		let validConfigurations = Object.keys(this.configurationSpecificData),
```
rosen-vladimirov: `_.keys` please

lib/services/livesync/livesync-service.ts
Mitko-Kerezov: I'll go ahead and extract it as a method, however calling it inside the `if` and respectively the `else` branch is an optimization. In the `if` the call is made using the original livesync information, whereas in the else multiple calls are made with configuration-specific livesync information

rosen-vladimirov:
:+1:  great work

PR #1447, opened 2016-06-15
Move TARGET_FRAMEWORK_IDENTIFIERS to constants

rosen-vladimirov:
Move TARGET_FRAMEWORK_IDENTIFIERS from $projectConstants (available in AppBuilder CLI only) to constants.
This way we can use it in common lib and all CLIs directly.


enchev:
Please check my comment here:
https://github.com/NativeScript/nativescript-cli/pull/1848#issuecomment-226118637

PR #1454, opened 2016-06-21
Add cli branch to config.json

TsvetanMilanov:
When executing the prepackage command we check if there are environment variables set by Jenkins for the sha of the commit (sha1) or for the branch name (BranchToBuild).


justcodebuilduser:
Can one of the admins verify this patch?

Mitko-Kerezov:
Seems legit 👍

PR #1467, opened 2016-07-06
Remove logic about storing configuration-specific keys

Mitko-Kerezov:
Instead of caching the old keys, do a deep diff between configuration-specific data and current project data. Save only configuration-specific data to configuration files.

In addition, if no configuration is specified, set property in base configuration file (.abproject) and delete keys in the other configuration files (.debug.abproject, .release.abproject).
This is accomplished by setting the property of all configuration-specific data objects in addition to the project data object to the same value.

Ping @rosen-vladimirov and @TsvetanMilanov 

Merge after https://github.com/telerik/mobile-cli-lib/pull/742 (although this PR doesn't depend on it)

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -407,6 +407,7 @@ export class Project extends ProjectBase implements Project.IProject {
 					});
 				} else {
 					this.$projectPropertiesService.updateProjectProperty(this.projectData, undefined, mode, normalizedPropertyName, propertyValues).wait();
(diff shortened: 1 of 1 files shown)


TsvetanMilanov:
run ci

TsvetanMilanov:
(Y)

TsvetanMilanov:
run ci

PR #1473, opened 2016-07-14
Fix plugin fetch from plugins directory

TsvetanMilanov:
When the user tries to fetch plugin from the plugins directory in the project we do not need to copy again the already existing plugin.

### DIFF
--- a/lib/services/plugins/cordova-project-plugins-service.ts
+++ b/lib/services/plugins/cordova-project-plugins-service.ts
@@ -321,11 +321,11 @@ export class CordovaProjectPluginsService extends NpmPluginsServiceBase implemen
 		}).future<IBasicPluginInformation>()();
 	}
 
-	protected fetchPluginBasicInformationCore(pathToInstalledPlugin: string, pluginData?: ILocalPluginData): IFuture<IBasicPluginInformation> {
+	protected fetchPluginBasicInformationCore(pathToInstalledPlugin: string, pluginData?: ILocalPluginData, options?: NpmPlugins.IFetchLocalPluginOptions): IFuture<IBasicPluginInformation> {
 		// We do not need to add the plugin to .abproject file because it will be sent with the plugins directory.
 		pluginData.addPluginToConfigFile = false;
 
-		return super.installLocalPlugin(pathToInstalledPlugin, pluginData);
+		return super.installLocalPlugin(pathToInstalledPlugin, pluginData, options);
 	}
 
 	private loadPluginsData(): IFuture<void> {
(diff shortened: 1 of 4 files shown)


lib/services/plugins/cordova-project-plugins-service.ts
```diff
@@ -321,11 +321,11 @@ export class CordovaProjectPluginsService extends NpmPluginsServiceBase implemen
 		}).future<IBasicPluginInformation>()();
 	}
 
-	protected fetchPluginBasicInformationCore(pathToInstalledPlugin: string, pluginData?: ILocalPluginData): IFuture<IBasicPluginInformation> {
+	protected fetchPluginBasicInformationCore(pathToInstalledPlugin: string, pluginData?: ILocalPluginData, options: NpmPlugins.IFetchLocalPluginOptions = { useOriginalPluginDirectory: false }): IFuture<IBasicPluginInformation> {
```
Mitko-Kerezov: You can set the `options` argument as non-mandatory only without setting a default value. The `installLocalPlugin` will set the default value if needed.

lib/services/plugins/nativescript-project-plugins-service.ts
```diff
@@ -220,7 +220,7 @@ export class NativeScriptProjectPluginsService extends NpmPluginsServiceBase imp
 		}).future<IBasicPluginInformation>()();
 	}
 
-	protected fetchPluginBasicInformationCore(pathToInstalledPlugin: string, pluginData?: ILocalPluginData): IFuture<IBasicPluginInformation> {
+	protected fetchPluginBasicInformationCore(pathToInstalledPlugin: string, pluginData?: ILocalPluginData, options: NpmPlugins.IFetchLocalPluginOptions = { useOriginalPluginDirectory: false }): IFuture<IBasicPluginInformation> {
```
Mitko-Kerezov: You can set the `options` argument as non-mandatory only without setting a default value. The `installLocalPlugin` will set the default value if needed.

lib/services/plugins/npm-plugins-service-base.ts
```diff
@@ -187,7 +192,7 @@ export abstract class NpmPluginsServiceBase implements IPluginsService {
 		return this.isLocalPath(pluginidentifier).wait() && (pluginIdentifierExtname === ".tgz" || pluginIdentifierExtname === ".gz");
 	}
 
-	protected fetchPluginBasicInformation(pluginIdentifier: string, failMessageMethodName: string, pluginData?: ILocalPluginData): IFuture<IBasicPluginInformation> {
+	protected fetchPluginBasicInformation(pluginIdentifier: string, failMessageMethodName: string, pluginData?: ILocalPluginData, options: NpmPlugins.IFetchLocalPluginOptions = { useOriginalPluginDirectory: false }): IFuture<IBasicPluginInformation> {
```
Mitko-Kerezov: Same remark for the `options` argument

Mitko-Kerezov:
👍 after addressing minor comments

PR #1474, opened 2016-07-14
Default to first configuration

Mitko-Kerezov:
Whenever running build/deploy without specifying any configuration default to the first (alphabetically sorted) configuration that the user has. If no configurations are present still default to debug though

Ping @rosen-vladimirov @TsvetanMilanov

PR #1475, opened 2016-07-14
Default to first configuration

Mitko-Kerezov:
Whenever running build/deploy without specifying any configuration default to the first (alphabetically sorted) configuration that the user has. If no configurations are present still default to debug though

Re-targeted from https://github.com/Icenium/icenium-cli/pull/1474


rosen-vladimirov:
:+1:

PR #1476, opened 2016-07-14
Fix images in html pages

TsvetanMilanov:
We need to use images in htmls as base64 strings because we serve only the html page without the images.


rosen-vladimirov:
:+1:

Mitko-Kerezov:
👍

PR #1479, opened 2016-07-15
Update osenv dependency

rosen-vladimirov:
Update osenv dependency in order to resolve issue on Mac OS X and Linux when HOME environment variable is not set.


Mitko-Kerezov:
👍

PR #1481, opened 2016-07-15
Default to lexicographically first configuration

Mitko-Kerezov:
Both when building for app and for companion. If no configurations present default to debug.


TsvetanMilanov:
:+1:

PR #1482, opened 2016-07-15
Search http://registry.npmjs.org for plugins

TsvetanMilanov:
When existing plugin name is passed to the fetch command it takes very long time to fetch the plugin. Slow fetch is expected only when keywords are passed to the fetch command. We can send request to http://registry.npmjs.org/[plugin-name] and see if there is registered plugin before using the slow npm find command.

### DIFF
--- a/lib/services/plugins/cordova-project-plugins-service.ts
+++ b/lib/services/plugins/cordova-project-plugins-service.ts
@@ -32,9 +32,10 @@ export class CordovaProjectPluginsService extends NpmPluginsServiceBase implemen
 		$project: Project.IProject,
 		$projectConstants: Project.IConstants,
 		$childProcess: IChildProcess,
+		$httpClient: Server.IHttpClient,
 		$hostInfo: IHostInfo,
 		$progressIndicator: IProgressIndicator) {
-		super($errors, $logger, $prompter, $fs, $project, $projectConstants, $childProcess, $hostInfo, $progressIndicator);
+		super($errors, $logger, $prompter, $fs, $project, $projectConstants, $childProcess, $httpClient, $hostInfo, $progressIndicator);
 	}
 
 	private get identifierToPlugin(): IDictionary<IPlugin> {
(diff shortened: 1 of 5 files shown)


lib/services/plugins/npm-plugins-service-base.ts
```diff
@@ -10,33 +10,43 @@ import temp = require("temp");
...
+
 			this.$progressIndicator.showProgressIndicator(npmFuture, 2000).wait();
 
 			let npmSearchResult = npmFuture.get();
 
 			if (npmSearchResult.stderr) {
-				this.$errors.failWithoutHelp(npmSearchResult.stderr);
+				// npm will write "npm WARN Building the local index for the first time, please be patient" to the stderr and if it is the only message on the stderr we should ignore it.
+				let splitError = npmSearchResult.stderr.split("\n");
+				if (splitError.length > 2 || !(splitError[0].indexOf("Building the local index for the first time") >= 0)) {
```
Mitko-Kerezov: If `splitError` is an empty array you'd get an error here

lib/services/plugins/cordova-project-plugins-service.ts
```diff
@@ -602,7 +603,11 @@ export class CordovaProjectPluginsService extends NpmPluginsServiceBase implemen
 			pluginName = pluginName.toLowerCase();
 			let isConsoleInteractive = helpers.isInteractive();
 			let allInstalledPlugins = this.getInstalledPluginsForConfiguration();
-			let installedPluginInstances = _.filter(allInstalledPlugins, (plugin: IPlugin) => plugin.data.Name.toLowerCase() === pluginName || plugin.data.Identifier.toLowerCase() === pluginName);
+			let installedPluginInstances = _(allInstalledPlugins)
+				.filter((plugin: IPlugin) => plugin.data.Name.toLowerCase() === pluginName || plugin.data.Identifier.toLowerCase() === pluginName)
+				.uniqBy((plugin: IPlugin) => plugin.data.Version)
```
rosen-vladimirov: why?

lib/services/plugins/npm-plugins-service-base.ts
rosen-vladimirov: instead of `!(splitError[0]...` you can use `|| splitError[0].indexOf("Building the local index for the first time") === -1)`

rosen-vladimirov:
:+1:

lib/services/plugins/cordova-project-plugins-service.ts
TsvetanMilanov: When we add plugin which is already added in more than one configurations with same version, we will get a warning `Plugin 'nl.x-services.plugins.socialsharing' is enabled with different versions in your project configurations.` because in `allInstalledPlugins ` there will be more than one instances of the plugin with the same version and that's why we need to filter the installed plugins by version. If the plugin is installed in more than one configurations with same version in the instance that is left after the filtering the `configurations` property will contain all the configurations of the plugin.

PR #1484, opened 2016-07-18
Revert Screen Builder 0.0.17

Mitko-Kerezov:
The new version currently does not work on OS X.
In addition the workflow for working with editablelists in the new version is somewhat obscured and a has diverged a bit from its UI counterpart.

Ping @rosen-vladimirov @TsvetanMilanov


rosen-vladimirov:
:+1: :100: :3000:

TsvetanMilanov:
:+1:

PR #1486, opened 2016-07-19
Fix plugin fetch for plugins with dependencies

TsvetanMilanov:
When we try to fetch plugin with dependencies npm 2 returns all installed dependencies on separate rows and the last row contains the plugin we want to fetch.


rosen-vladimirov:
:+1:

PR #1487, opened 2016-07-19
Fix failing plugin fetch tests

TsvetanMilanov:
Wehn we install the plugin with npm to Temp directory we need to check if the plugin is installed in node_modules in the Temp folder both for NativeScript and for Cordova projects.


rosen-vladimirov:
:+1:

PR #1490, opened 2016-07-19
Fix prompter memory leak

TsvetanMilanov:
One of our dependencies mute-stream has function pipe which calls the pipe of the Node js Stream and if the pipe is called with same destination multiple times the event listeners added to the stream will cause memory leak.
We need to check the event listeners of the process.stdout and clean them before we call pipe of MuteStream.


rosen-vladimirov:
:+1:

rosen-vladimirov:
common lib SHA will be updated with another PR

PR #1491, opened 2016-07-20
Fix starting of NativeScript applications

rosen-vladimirov:
Fix starting of NativeScript applications by passing correct startPackageActivityName. Use the framework from projectData in case it exists in order to determine the startPackageActivity.

http://teampulse.telerik.com/view#item/318259


Mitko-Kerezov:
👍 and don't forget to update the SHA once you merge in the common lib

PR #1493, opened 2016-07-20
Respect default option when adding NS plugin

TsvetanMilanov:
We need to respect the --default option when adding NS plugin and also we need to change the sorting of the plugin versions when we get the latest plugin verison.

### DIFF
--- a/lib/services/plugins/nativescript-project-plugins-service.ts
+++ b/lib/services/plugins/nativescript-project-plugins-service.ts
@@ -30,9 +30,10 @@ export class NativeScriptProjectPluginsService extends NpmPluginsServiceBase imp
 		$projectConstants: Project.IConstants,
 		$childProcess: IChildProcess,
 		$httpClient: Server.IHttpClient,
+		$options: IOptions,
 		$hostInfo: IHostInfo,
 		$progressIndicator: IProgressIndicator) {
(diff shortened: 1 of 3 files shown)


lib/services/plugins/npm-plugins-service-base.ts
```diff
@@ -322,6 +323,10 @@ export abstract class NpmPluginsServiceBase implements IPluginsService {
 		}).future<boolean>()();
 	}
 
+	protected getDefauldPluginVersion(plugin: IPlugin): string {
```
rosen-vladimirov: getDefaul**t**PluginVersion maybe

lib/services/plugins/nativescript-project-plugins-service.ts
```diff
@@ -477,8 +477,7 @@ export class NativeScriptProjectPluginsService extends NpmPluginsServiceBase imp
 			let allMarketplacePlugins = this.getMarketplacePlugins().wait();
 			let marketPlacePlugins: IPlugin[] = _.filter(allMarketplacePlugins, pl => pl.data.Identifier.toLowerCase() === pluginName.toLowerCase());
 			if (marketPlacePlugins && marketPlacePlugins.length) {
-				let selectedPlugin = version === "latest" ? _.last(_.sortBy(marketPlacePlugins, pl => pl.data.Version)) :
```
rosen-vladimirov: we need the `latest` back as we can write
```
$ appbuilder plugin add nativescript-push-notifications@latest
```

PR #1495, opened 2016-07-20
Check if app is installed when getting the parsed plist

TsvetanMilanov:
If the application is not installed and we try to get the plist content of it we will fail.


TsvetanMilanov:
common lib SHA will be updated with another PR

PR #1497, opened 2016-07-22
Fix adding CorePlugins to specific configuration

rosen-vladimirov:
- Fix `$ appbuilder prop set CorePlugins <values> --debug` not setting values to specific configuration.
- Fix `$ appbuilder plugin add console --release` not moving all CorePlugins to all available configurations.
- Fix possible adding of multiple plugins with same name in a single configuration (missing `uniq` in the code).

When change in CorePlugins is applied, save all possible .abproject files as all of them might be affected.
Extract the `updateProjectProperty` functionality from `updateProjectPropertyAndSave` method to a separate public method in order to keep the current logging for `plugin add` command.

Fixes http://teampulse.telerik.com/view#item/317915

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -402,17 +402,12 @@ export class Project extends ProjectBase implements Project.IProject {
 		return this.getConfigurationsSpecifiedByUser()[0] || _.first(this.getAllConfigurationsNames().sort()) || Configurations.Debug;
 	}
 
-	public updateProjectPropertyAndSave(mode: string, propertyName: string, propertyValues: string[]): IFuture<void> {
+	public updateProjectProperty(mode: string, propertyName: string, propertyValues: string[], configurations?: string[]): IFuture<void> {
 		return (() => {
-			this.ensureProject();
-			let projectConfigurations = this.getConfigurationsSpecifiedByUser();
-			let normalizedPropertyName = this.$projectPropertiesService.normalizePropertyName(propertyName, this.projectData);
-			if (normalizedPropertyName === this.$projectConstants.APPIDENTIFIER_PROPERTY_NAME) {
-				this.$jsonSchemaValidator.validatePropertyUsingBuildSchema(normalizedPropertyName, propertyValues[0]);
-			}
+			let { normalizedPropertyName, projectConfigurations } = this.validateUpdatePropertyInfo(propertyName, propertyValues, configurations);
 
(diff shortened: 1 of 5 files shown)


lib/project.ts
```diff
@@ -423,8 +423,19 @@ export class Project extends ProjectBase implements Project.IProject {
 					_.each(this.configurationSpecificData, data => this.$projectPropertiesService.updateProjectProperty(data, undefined, mode, normalizedPropertyName, propertyValues).wait());
 				}
 			}
+		}).future<void>()();
+	}
+
+	public updateProjectPropertyAndSave(mode: string, propertyName: string, propertyValues: string[], configurations?: string[]): IFuture<void> {
+		return (() => {
+			this.ensureProject();
```
Mitko-Kerezov: You can extract the beginning of `updateProjectPropertyAndSave` and `updateProjectProperty` into a separate private method, which returns information about `projectConfigurations` and `normalizedPropertyName` in order to avoid code duplication

lib/services/plugins/cordova-project-plugins-service.ts
```diff
@@ -438,15 +439,20 @@ export class CordovaProjectPluginsService extends NpmPluginsServiceBase implemen
...
-			let versionString = this.isMarketplacePlugin(plugin) ? ` with version ${version}` : "";
+
+			let versionString = this.isMarketplacePlugin(plugin) ? ` with version ${version}` : "",
+				successMessageForConfigSuffix = ` for ${configuration} configuration${versionString}`,
+				successMessage = `Plugin ${pluginName} was successfully added${configuration ? successMessageForConfigSuffix : versionString}.`;
+
 			if (configuration) {
-				this.$project.saveProject(this.$project.getProjectDir().wait(), [configuration]).wait();
-				this.$logger.out(`Plugin ${pluginName} was successfully added for ${configuration} configuration${versionString}.`);
+				this.$project.updateProjectProperty("set", CordovaProjectPluginsService.CORE_PLUGINS_PROPERTY_NAME, newCorePlugins, [ configuration ]).wait();
```
Mitko-Kerezov: We should definately extract all the project property modes into constants. In another PR, though

Mitko-Kerezov:
LGTM :+1:

TsvetanMilanov:
:+1:

tytodorov:
👍

PR #1500, opened 2016-07-26
Update to latest common lib API for mapAbstractToTcpPort

rosen-vladimirov:
Pass framework as required by new API in common lib when mapping abstract to TCP port.


TsvetanMilanov:
:+1:

PR #1501, opened 2016-07-28
Enable AM LiveSync feature for {N} apps.

ADjenkov:
General changes:
 - Extending `appmanager-service` to support {N} apps.
 - __appmanager livesync__ command should support only iOS and Android {N} apps.

Nativescript plugins service changes (`nativescript-project-plugins-service`):
 -  Introduced "Advanced Plugins" section when showing available plugins to the user. _I found the name "Featured Plugins" a lot more appropriate but i guess it should be consistent with Cordova?_
   
     I've decided to introduce `getFeaturedNpmPackages` that can handle more than one plugin that needs to show up under "Advanced Plugins". But I could remove it and `concat` manually the AM LiveSync plugin as an 'Advanced Plugin' since for now we only have one plugin to display there. 
 - Skipping variables configuration when adding AM LiveSync plugin since they are not relevant to the user and they will be populated on the server.

**Important note:** AM LiveSync plugin for {N} is available in NPM, unlike AM LiveSync plugin for Cordova. That's why we are fetching all the plugin info from NPM as a regular package not mocking it like Cordova.

### DIFF
--- a/lib/commands/appmanager-livesync.ts
+++ b/lib/commands/appmanager-livesync.ts
@@ -1,7 +1,10 @@
-﻿export class AppManagerLiveSyncCommand implements ICommand {
+﻿import { TARGET_FRAMEWORK_IDENTIFIERS } from "../common/constants";
+
+export class AppManagerLiveSyncCommand implements ICommand {
 	private static ALL_PLATFORMS_OPTION = "All platforms";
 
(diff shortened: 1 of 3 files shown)


rosen-vladimirov:
:+1:

ADjenkov:
run ci

PR #1503, opened 2016-08-01
Fix publish of the CLI

rosen-vladimirov:
Fix publishing of the CLI as currently the published package does not have it's scripts section. The reason is that we pass `--ignore-scripts` to `npm publish` command. This removes all scripts from the published package.
After removing `--ignore-scripts` it turned out the package cannot be published. The problem is that when `npm publish <tgz>` is used, the package is added to local npm cache. After that it's package.json is extracted from the cached .tgz file and modified with additional metadata. At this point npm calls the prepublish script, but it fails to find our prepublish.js as it's not extracted in the cache (only the package.json is extracted).
So in order to fix this, remove the `prepublish` script in the `grunt pack` step.

I've not removed the prepublish script from the package.json, so if anyone uses `npm pack` locally or `npm publish` directly, the correct resources will be downloaded.
> NOTE: Calling `grunt pack` locally will modify your package.json. DO NOT commit these changes.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,7 +4,7 @@ var now = new Date().toISOString();
 
 function shallowCopy(obj) {
 	var result = {};
-	Object.keys(obj).forEach(function(key) {
+	Object.keys(obj).forEach(function (key) {
 		result[key] = obj[key];
 	});
 	return result;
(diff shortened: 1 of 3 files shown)


teobugslayer:
@rosen-vladimirov - can you add a big, bold, blinking, red warning in `grunt pack` not to commit the changed file?

rosen-vladimirov:
@teobugslayer done - I've added a read message at the end of `grunt pack` command.

TsvetanMilanov:
:+1:

PR #1504, opened 2016-08-02
Download build results from s3

TsvetanMilanov:
If we want to download the build results directly from s3 we need to send one additional property in the request - AcceptResults, with value - Url;LocalPath. In the response the link to the result will be in the FullPath property and the Format will be Url. In case we can't download the build results from s3 (wp8) we need to use the LocalPath method.

### DIFF
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -201,7 +201,7 @@ declare module Project {
 	}
 
 	interface IBuildService {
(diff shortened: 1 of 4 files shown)


rosen-vladimirov:
What about the QR code generation? Should we do anything special there ?

rosen-vladimirov:
:+1:

TsvetanMilanov:
run ci

PR #1506, opened 2016-08-09
Add support for installing npm modules and @types

rosen-vladimirov:
Add support for installing npm dependencies. Provide two methods for public API:
 - install - installs everything from package.json. In case a specific dependency must be installed, it's described as second argument with its name, version and should `@types` be installed.
 - uninstall - uninstalls the specified dependency from both node_modules and package.json. The call will also remove `@types/<dependency>` entry from devDependencies.

On each build related command in NativeScript TypeScript project, CLI will call `npm install` in order to verify all required dependencies are installed.
On plugin add - try adding `@types/<dependency>` in case it exists and `--no-types` is not passed on the terminal.
On plugin remove - remove both the dependency and `@types/<dependency>` entries.
http://teampulse.telerik.com/view#item/317268

Remove typings/tns-core-modules after user's confirmation as it just conflicts with our new code: http://teampulse.telerik.com/view#item/317583

Do not create typings/tns-core-modules when migrating between {N} versions, Simplify migration logic http://teampulse.telerik.com/view#item/319382
Renamed platformMigrator to projectMigrationService.

Introduce public API in common lib for installing dependencies from Proton: http://teampulse.telerik.com/view#item/318796

Fix:
 - CLI is unable to install scoped dependencies (`@angular/core` for example) as:
   - we assume everything after `@` is the version
   - we encode the name, so `@angular/core` becomes `%40angular%2Fcore`, but registry.npmjs.org does not understand this, only `/` must be encoded.

Still to do (in separate PRs)
* Add android17.d.ts and ios.d.ts to the project and reference them in .abreferences.d.ts
* Add unit tests

### DIFF
--- a/lib/services/plugins/nativescript-project-plugins-service.ts
+++ b/lib/services/plugins/nativescript-project-plugins-service.ts
@@ -15,14 +15,15 @@ export class NativeScriptProjectPluginsService extends NpmPluginsServiceBase imp
 	private static NPM_SEARCH_URL = "http://npmsearch.com";
 	private static HEADERS = ["NPM Packages", "NPM NativeScript Plugins", "Marketplace Plugins", "Advanced Plugins"];
 	private static DEFAULT_NUMBER_OF_NPM_PACKAGES = 10;
-	private static NPM_REGISTRY_URL = "https://registry.npmjs.org";
 	private static NATIVESCRIPT_LIVEPATCH_PLUGIN_ID = "nativescript-plugin-livepatch";
 
 	private featuredNpmPackages = [NativeScriptProjectPluginsService.NATIVESCRIPT_LIVEPATCH_PLUGIN_ID];
 	private marketplacePlugins: IPlugin[];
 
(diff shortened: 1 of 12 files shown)


TsvetanMilanov:
:+1:

lib/services/plugins/nativescript-project-plugins-service.ts
```diff
@@ -175,7 +186,15 @@ export class NativeScriptProjectPluginsService extends NpmPluginsServiceBase imp
 
 	public getPluginBasicInformation(pluginName: string): IFuture<IBasicPluginInformation> {
 		return ((): IBasicPluginInformation => {
-			let [name, version] = pluginName.split("@");
+			let name: string, version: string;
+			let scopedDependencyRegex = /^(@.+?)(?:@(.+?))?$/;
```
Mitko-Kerezov: The combination of `+?` is equivalent to `*`, not a merge-stopper but the two occurrences can be replaced and therefore shortened

Mitko-Kerezov:
LGTM :+1:

PR #1510, opened 2016-08-10
Start App with identifier returned from the server

TsvetanMilanov:
Since the app identifier can be changed in AndroidManifest.xml and config.xml the server needs to return the identifier which is used to build the application. That's why we need to use the app identifier from the server instead of the one from the .abproject.

### DIFF
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -1316,7 +1316,17 @@ interface IDateProvider {
 	getCurrentDate(): Date;
 }
 
+/**
+ * Describes information about application package.
+ */
 interface IApplicationInformation {
+	/**
+	 * The name of the package file.
+	 */
 	packageName: string;
+
+	/**
+	 * The identifier of the application.
+	 */
 	appIdentifier: string;
 }
--- a/lib/declarations.d.ts
+++ b/lib/declarations.d.ts
@@ -40,6 +40,9 @@ declare module Server {
 		format: string;
 		url: string;
 		fileName: string;
+		key?: string;
+		value?: string;
+		architecture?: string;
 	}
 
 	interface IBuildResult {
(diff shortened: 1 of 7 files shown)


TsvetanMilanov:
run ci

TsvetanMilanov:
run ci

.vscode/launch.json
```diff
@@ -13,9 +13,9 @@
 			// Automatically stop program after launch.
 			"stopOnEntry": true,
 			// Command line arguments passed to the program.
-			"args": [],
+			"args": ["emulate", "android"],
```
Mitko-Kerezov: I believe this file is committed by mistake

lib/services/build.ts
```diff
@@ -444,14 +453,19 @@ export class BuildService implements Project.IBuildService {
 				device: device
 			}).wait();
 
-			let result = _.filter(buildResult, (def: Server.IPackageDef) => !def.disposition || def.disposition === "BuildResult")[0].localFile;
-			return result;
-		}).future<string>()();
+			let packageName = _.filter(buildResult, (def: Server.IPackageDef) => !def.disposition || def.disposition === "BuildResult")[0].localFile;
+			let metadata = _.filter(buildResult, (def: Server.IPackageDef) => !def.disposition || (def.disposition === "BuildResultMetadata" && def.key === "AppIdentifier"))[0];
```
Mitko-Kerezov: If by chance the first `IPackageDef` doesn't have `disposition` set you'd end up with `packageName` and `metadata` being the same. That may or may not be a problem - what do you think @TsvetanMilanov

Mitko-Kerezov: I do believe that this would be better:
```TypeScript
let metadata = _.filter(buildResult, (def: Server.IPackageDef) => def.disposition === "BuildResultMetadata" && def.key === "AppIdentifier")[0];
```

Since you're only actually interested in the `AppIdentifier` itself.

lib/declarations.d.ts
```diff
@@ -1312,3 +1315,8 @@ declare module NpmPlugins {
 interface IDateProvider {
 	getCurrentDate(): Date;
 }
+
+interface IApplicationInformation {
```
Mitko-Kerezov: Some docstrings for this would be nice

Mitko-Kerezov:
👍 great work on this!

PR #1514, opened 2016-08-16
Provide a way to find scoped dependencies

TsvetanMilanov:
We need to have a way to find scoped dependencies (e.g. @angular/core).
Currently the flow is:
Call https://www.npmjs.com/search?q={query}
If that fails:
If the query starts with @, we know npm search will fail, so try to get the package details directly from http://registry.npmjs.org. It will only work if the name is exact match, but it's better than nothing.
If the query does not start with @, fall back to npm search.


rosen-vladimirov:
:+1:  after rebasing

PR #1515, opened 2016-08-17
Fix resolve livesync service

TsvetanMilanov:
After `platformSpecificLiveSyncServices` was renamed to `deviceSpecificLiveSyncServices` when we try to resolve livesync service we need to use `deviceSpecificLiveSyncServices` instead of the empty `platformSpecificLiveSyncServices`.


Mitko-Kerezov:
👍

PR #1516, opened 2016-08-19
Get the correct app id

TsvetanMilanov:
Cherry-pick the following commits from master:

#### Start App with identifier returned from the server
Since the app identifier can be changed in AndroidManifest.xml and config.xml the server needs to return the identifier which is used to build the application. That's why we need to use the app identifier from the server instead of the one from the .abproject.

#### Get the correct app id
We need to get the correct app id which will be used to build the application on our server when we execute operations for Android.

### DIFF
--- a/lib/commands/debug.ts
+++ b/lib/commands/debug.ts
@@ -3,6 +3,7 @@ import Future = require("fibers/future");
 
 export class DebugCommand implements ICommand {
 	private debuggerPath: string;
+	protected platform: string;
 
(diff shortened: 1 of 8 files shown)


Mitko-Kerezov:
Cherry pick from https://github.com/Icenium/icenium-cli/pull/1510

Mitko-Kerezov:
👍

tytodorov:
👍

TsvetanMilanov:
run ci

PR #1517, opened 2016-08-19
Fix unexpected token o

TsvetanMilanov:
The package.json information returned from the npm service is an object and when we construct the NS plugin data we expect it to be a string and JSON.parse() fails.


Mitko-Kerezov:
LGTM :+1:

nirinchev:
👍

TsvetanMilanov:
run ci

PR #1518, opened 2016-08-22
Delete typings dir if it is empty

TsvetanMilanov:
After we migrate TS projects we should delete the typings directory if it is empty after we remove the tns-core-modules directory.


.vscode/launch.json
```diff
@@ -13,9 +13,9 @@
 			// Automatically stop program after launch.
 			"stopOnEntry": true,
 			// Command line arguments passed to the program.
-			"args": [],
+			"args": ["mobileframework", "set", "2.2.0"],
```
rosen-vladimirov: revert please

rosen-vladimirov:
:+1:

PR #1519, opened 2016-08-22
Fix error when getting android id

TsvetanMilanov:
When we have only AndroidManifest.xml or config.xml in the project and the app id in it is not changed we do not need to throw exception for app id mismatch.


rosen-vladimirov:
:+1:  after updating sha of common lib and green build

PR #1520, opened 2016-08-22
Revert {N} plugin variables default value key

Mitko-Kerezov:
{N} plugin variables' default values are set through the key `defaultValue` instead of the key `default`


TsvetanMilanov:
:+1:

PR #1522, opened 2016-08-23
Fix pathToPlugin.indexOf is not a function

TsvetanMilanov:
When we try to fetch plugin for Cordova project from github we should not treat it as local plugin.

### DIFF
--- a/lib/services/plugins/npm-plugins-service-base.ts
+++ b/lib/services/plugins/npm-plugins-service-base.ts
@@ -349,9 +349,16 @@ export abstract class NpmPluginsServiceBase implements IPluginsService {
 				suppressMessage: !suppressMessage
 			};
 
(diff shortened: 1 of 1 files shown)


lib/services/plugins/npm-plugins-service-base.ts
```diff
@@ -349,9 +349,16 @@ export abstract class NpmPluginsServiceBase implements IPluginsService {
 				suppressMessage: !suppressMessage
 			};
 
-			options.originalPluginDirectory = pluginLocalPathExists && pluginLocalPath;
+			let pluginId = pluginIdentifier;
```
rosen-vladimirov: Why do you need the new variable. As far as I see, `pluginIdentifier` is not used anywhere else, so you can just set it to `pluginLocalPath` in the `if` body

rosen-vladimirov:
:+1:

PR #1523, opened 2016-08-23
Implement force upgrade logic for projects using cordova android 4.0.2

Mitko-Kerezov:
Update projects that are using Cordova 4.0.0 (Android 4.0.2) to use the Cordova 4.1.1 and inform the user for this action.
TP reference [Update project's that are using Cordova 4.0.0 (Android 4.0.2) to use the Cordova 4.1.1](http://teampulse.telerik.com/view#item/310575)

### DIFF
--- a/lib/project/cordova-project.ts
+++ b/lib/project/cordova-project.ts
@@ -9,6 +9,7 @@ import Future = require("fibers/future");
 export class CordovaProject extends FrameworkProjectBase implements Project.IFrameworkProject {
 	private static WP8_DEFAULT_PACKAGE_IDENTITY_NAME_PREFIX = "1234Telerik";
 	private static WP8_DEFAULT_WP8_WINDOWS_PUBLISHER_NAME = "CN=Telerik";
+	private static WHITESPACE_REGEX = /\s/g;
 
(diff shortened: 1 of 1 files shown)


lib/project/cordova-project.ts
```diff
@@ -168,10 +169,13 @@ export class CordovaProject extends FrameworkProjectBase implements Project.IFra
 
 	private ensureCordovaJs(platform: string, projectDir: string, frameworkVersion: string): IFuture<void> {
 		return (() => {
-			let cordovaJsFileName = path.join(projectDir, util.format("cordova.%s.js", platform).toLowerCase());
-			if (!this.$fs.exists(cordovaJsFileName).wait()) {
+			let cordovaJsFileName = path.join(projectDir, `cordova.${platform.toLowerCase()}.js`),
```
TsvetanMilanov: Can you change the name of `cordovaJsFileName ` to `cordovaJsFilePath` because this variable contains the path to the file not the name of it.

nirinchev:
While using Regex is 6 times slower than using the magic `IgnoreWhiteSpaceComparer`, I think we can live with the performance hit for now 👍

TsvetanMilanov:
:+1:

PR #1524, opened 2016-08-24
Fix add and fetch scoped dependencies

TsvetanMilanov:
When we fetch scoped dependencies we need to get the correct directory in which the dependency is installed because it will be in `node_modules/@<scope>/<dependency>` instead of `node_modules/<dependency>`.


rosen-vladimirov:
:+1:

PR #1525, opened 2016-08-25
Support noEmitOnError parameter and execute hooks in process

rosen-vladimirov:
Previously there were some cases when TypeScript was unable to generate .js files. After adding the `noEmitOnError` parameter to tsconfig, TypeScript will not generate .js files ONLY when it is set to true.
Currently our TypeScriptService does not handle this case as it thinks that the .js files are not generated in some cases. Respect the parameter and always show the output of the compiler to the user.

Also make sure all of our hooks are executed in the current process, so the error messages from them will be correctly shown to the user. This requires adding `esprima` as dependency as its used in mobile-cli-lib code.

### DIFF
--- a/lib/hooks/before-build.js
+++ b/lib/hooks/before-build.js
@@ -1,1 +1,1 @@
-require("./typescript-compilation.js");
+module.exports = require("./typescript-compilation.js");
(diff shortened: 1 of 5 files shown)


Mitko-Kerezov:
LGTM :+1:

PR #1532, opened 2016-08-30
Regenerate server api

Mitko-Kerezov:
Regenerate server api and fix all compilation errors:
* Use endpoint to upload package from URI when performing `appmanager upload`
* Use endpoint to upload package from URI when performing `appstore upload`
* Skip passing `checkUpgradability` argument which is no longer supported when calling `getApplication`
* Extract constants file for storing AppBuilder related constants
* Pass `user-settings.xml` as file name when getting or saving user settings

TP Ref: [Cannot publish to AppManager](http://teampulse.telerik.com/view#item/320571)

### DIFF
--- a/lib/services/appmanager-service.ts
+++ b/lib/services/appmanager-service.ts
@@ -76,7 +76,9 @@ class AppManagerService implements IAppManagerService {
 				this.$logger.warn("You have not set the --publish switch. Your users will not receive a push notification.");
 			}
 
-			let uploadedAppData: Server.UploadedAppData = this.$server.tam.uploadApplication1(projectName, projectName, projectPath, publishSettings).wait();
+			let uploadedAppData: Server.UploadedAppData = mobilePlatform.toLowerCase() === this.$devicePlatformsConstants.WP8.toLowerCase() ?
+				this.$server.tam.uploadApplication(projectName, projectName, projectPath, publishSettings).wait() :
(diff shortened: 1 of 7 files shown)


lib/services/appmanager-service.ts
```diff
@@ -76,7 +76,7 @@ class AppManagerService implements IAppManagerService {
 				this.$logger.warn("You have not set the --publish switch. Your users will not receive a push notification.");
 			}
 
-			let uploadedAppData: Server.UploadedAppData = this.$server.tam.uploadApplication1(projectName, projectName, projectPath, publishSettings).wait();
+			let uploadedAppData: Server.UploadedAppData = this.$server.tam.uploadApplicationFromUri(projectName, projectName, <any>projectPath, publishSettings).wait();
```
TsvetanMilanov: Since this `<any>projectPath` is working we need to check why the `uploadApplicationFromUri` is generated with `packageUri: Server.Uri` instead of `packageUri: string`

TsvetanMilanov:
:+1:

lib/services/appmanager-service.ts
rosen-vladimirov: this is expected, swagger respects the original type of the property

rosen-vladimirov: but this code will be called for Windows Phone as well... and it will fail

Mitko-Kerezov:
@TsvetanMilanov comment addressed, however we'd need to merge https://github.com/Icenium/Ice/pull/6190 before merging this

PR #1533, opened 2016-08-30
Print the device log for specific duration

TsvetanMilanov:
Since the device log command does not end unless the process is killed by the user we need to provide a way to execute it for specific duration.
This way it can be used in tests without blocking them.

Common reference https://github.com/telerik/mobile-cli-lib/pull/792

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,4 +6,6 @@
 		"**/*.js": { "when": "$(basename).ts"},
 		"**/*.js.map": true
 	}
+,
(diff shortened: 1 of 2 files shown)


TsvetanMilanov:
run ci

TsvetanMilanov:
run ci

TsvetanMilanov:
run ci

TsvetanMilanov:
run ci

TsvetanMilanov:
Closing this because this problem is out of our scope for now.

PR #1534, opened 2016-08-30
Respect configurations on appmanager/appstore upload

Mitko-Kerezov:
Expose a way to default to a desired configuration (if present among the user configurations) throught the `$project` dependency.
Set project configuration as:
* What the user has defined as `--config`
* `release` (if present among the user's configurations)
* The lexicographically first configuration the user has
* `Debug`

In that order.

Mind that this is a project configuration as opposed to build configuration and even in the unlikely case that we end up building in `Debug`, the build configuration will always be `release` thus preventing any unwanted behavior.

Fixes [appmanager and appstore upload do not respect --config ](http://teampulse.telerik.com/view#item/318698)

Ping @rosen-vladimirov @TsvetanMilanov

### DIFF
--- a/lib/project.ts
+++ b/lib/project.ts
@@ -398,8 +398,13 @@ export class Project extends ProjectBase implements Project.IProject {
 		return _.keys(this.configurationSpecificData);
 	}
 
-	public getProjectConfiguration(): string {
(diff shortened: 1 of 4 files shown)


TsvetanMilanov:
:+1:

PR #1535, opened 2016-08-30
Separate .d.ts files from .ts files.

TsvetanMilanov:
When we get the TS files data of the project we don't need the .d.ts files in the typeScriptFiles property.
Also when we add the TS files to tscommand.txt we need to add quotation marks to their names because some of them may contain spaces.


rosen-vladimirov:
:+1:

PR #1537, opened 2016-08-31
Install all npm dependencies in NS TS projects

TsvetanMilanov:
When we create NS TS project we need to install all npm dependencies because if we don't have the tns-core-modules installed we will not add theirs .d.ts in .abreferences file untill the project is built.
Also we need to install the dependencies on every project related command because we must be sure that we are working with correct content of node_modules.

### DIFF
--- a/lib/project/cordova-project.ts
+++ b/lib/project/cordova-project.ts
@@ -167,6 +167,10 @@ export class CordovaProject extends FrameworkProjectBase implements Project.IFra
 		}).future<void>()();
 	}
 
+	public ensureProject(projectDir: string): IFuture<void> {
+		return Future.fromResult();
+	}
+
 	private ensureCordovaJs(platform: string, projectDir: string, frameworkVersion: string): IFuture<void> {
 		return (() => {
 			let cordovaJsFilePath = path.join(projectDir, `cordova.${platform.toLowerCase()}.js`),
--- a/lib/project/nativescript-project.ts
+++ b/lib/project/nativescript-project.ts
@@ -16,12 +16,14 @@ export class NativeScriptProject extends FrameworkProjectBase implements Project
 		private $injector: IInjector,
 		private $nativeScriptProjectCapabilities: Project.ICapabilities,
 		private $dateProvider: IDateProvider,
+		private $typeScriptService: ITypeScriptService,
+		private $npmService: INpmService,
 		$errors: IErrors,
 		$fs: IFileSystem,
 		$jsonSchemaValidator: IJsonSchemaValidator,
 		$logger: ILogger,
-		$resources: IResourceLoader,
-		$options: IOptions) {
+		$options: IOptions,
+		$resources: IResourceLoader) {
 		super($logger, $fs, $resources, $errors, $jsonSchemaValidator, $options);
 	}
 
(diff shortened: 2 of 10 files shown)


lib/project/nativescript-project.ts
```diff
@@ -6,7 +6,7 @@ import {FrameworkProjectBase} from "./framework-project-base";
 import Future = require("fibers/future");
 temp.track();
 
-export class NativeScriptProject extends FrameworkProjectBase implements Project.IFrameworkProject {
+export class NativeScriptProject extends FrameworkProjectBase implements Project.IFrameworkProject, Project.IFrameworkProjectBase {
```
rosen-vladimirov: why? the IFrameworkProject already extends IFrameworkProjectBase

lib/project/cordova-project.ts
```diff
@@ -6,7 +6,7 @@ import helpers = require("./../common/helpers");
 import semver = require("semver");
 import Future = require("fibers/future");
 
-export class CordovaProject extends FrameworkProjectBase implements Project.IFrameworkProject {
+export class CordovaProject extends FrameworkProjectBase implements Project.IFrameworkProject, Project.IFrameworkProjectBase {
```
rosen-vladimirov: same as for NativeScript project

lib/project/nativescript-project.ts
```diff
@@ -182,6 +184,18 @@ export class NativeScriptProject extends FrameworkProjectBase implements Project
...
 		}).future<void>()();
 	}
+
+	public ensureProject(projectDir: string): IFuture<void> {
+		return (() => {
+			if (this.$typeScriptService.isTypeScriptProject(projectDir).wait()) {
+				try {
+					this.$npmService.install(projectDir).wait();
+				} catch (err) {
+					this.$logger.trace(`Filed to install all npm dependencies in the project. Error: ${err}.`);
```
rosen-vladimirov: Filed -> Failed

lib/project/cordova-project.ts
```diff
@@ -25,8 +25,8 @@ export class CordovaProject extends FrameworkProjectBase implements Project.IFra
 		$fs: IFileSystem,
 		$jsonSchemaValidator: IJsonSchemaValidator,
 		$logger: ILogger,
-		$options: IOptions,
-		$resources: IResourceLoader) {
+		$resources: IResourceLoader,
```
rosen-vladimirov: why is this change?

rosen-vladimirov:
:+1:

lib/project/cordova-project.ts
TsvetanMilanov: I wanted the order here to be the same as the order in `NativeScriptProject`. This way it will be easier to edit if we add new parameter in the constructor of `FrameworkProjectBase` :smile:

TsvetanMilanov:
run ci

lib/project/nativescript-project.ts
```diff
@@ -182,6 +184,18 @@ export class NativeScriptProject extends FrameworkProjectBase implements Project
...
 		}).future<void>()();
 	}
+
+	public ensureProject(projectDir: string): IFuture<void> {
+		return (() => {
+			if (this.$typeScriptService.isTypeScriptProject(projectDir).wait()) {
+				try {
+					this.$npmService.install(projectDir).wait();
+				} catch (err) {
+					this.$logger.trace(`Failed to install all npm dependencies in the project. Error: ${err}.`);
```
teobugslayer: If there is an exception, can we state that the project has all the required files? Shouldn't we rethrow the exception?

TsvetanMilanov: I think it is better to display warning here instead of rethrow the exception because if the npm install fails we can continue with the command execution with the current content of the node_modules in the project.

PR #1539, opened 2016-09-01
Fix get Android app id check

TsvetanMilanov:
When we compare the app identifiers in AndroidManifest.xml and in config.xml in case only one app identifier is changed we need to check if it is changed to a value which is different from the one in .abproject file.


Mitko-Kerezov:
👍

rosen-vladimirov:
:+1: 
As the builds are running against our sit environment, unit tests cannot pass. Real result is here: http://bpc15:8080/view/CLI/job/ICE_CLI_PR/2018/

ndinev:
:+1:

PR #1545, opened 2016-09-08
Fix adding missing cordova.<platform>.js file to project

rosen-vladimirov:
Before build CLI tries to ensure all `cordova.<platform>.js` files are part of the project.
When any of them is missing, CLI should extract it from the resources and add it to the project. Instead of adding it, CLI fails with error:
`Error: ENOENT: no such file or directory, open '<project dir>\cordova.<platform>.js'`
The reason is incorrect code in CLI that tries to verify if the content of the file in the project (no matter that it does not exist) is the same as the one in the resources of CLI.
Fixes http://teampulse.telerik.com/view#item/321100


TsvetanMilanov:
:+1:

Mitko-Kerezov:
LGTM :+1:

PR #1546, opened 2016-09-09
Fix get plugin author

TsvetanMilanov:
When we get the author of the plugin first we need to check if there is one.


Mitko-Kerezov:
👍

PR #1547, opened 2016-09-09
Update init command description

TsvetanMilanov:
Update the description of the init command because we can create AppBuilder project from Ionic project.


Mitko-Kerezov:
Looks okay 👍

PR #1549, opened 2016-09-09
Fix NS Companion app livesync

TsvetanMilanov:
When we create the Android NS companion app device app data we need to pass framework NativeScrit instead of Cordova.


TsvetanMilanov:
run ci

PR #1552, opened 2016-09-12
Set justlaunch flag when executing livesync

TsvetanMilanov:
When we execute livesync on iOS simulator we need to set the justlaunch option to true if there is no watch flag because if we don't set it the command will not release the console.

### DIFF
--- a/lib/commands/live-sync.ts
+++ b/lib/commands/live-sync.ts
@@ -1,63 +1,99 @@
 import { EnsureProjectCommand } from "./ensure-project-command";
+import Future = require("fibers/future");
 
-class LiveSyncDevicesCommand extends EnsureProjectCommand {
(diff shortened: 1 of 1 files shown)


lib/commands/live-sync.ts
```diff
@@ -1,63 +1,99 @@
 import { EnsureProjectCommand } from "./ensure-project-command";
+import Future = require("fibers/future");
 
-class LiveSyncDevicesCommand extends EnsureProjectCommand {
-	constructor(private $liveSyncService: ILiveSyncService,
+class LiveSyncCommandBase extends EnsureProjectCommand {
+	constructor(protected $liveSyncService: ILiveSyncService,
+		private $options: ICommonOptions,
```
rosen-vladimirov: as the code is in `icenium-cli` this should be `IOptions`

PR #1558, opened 2016-09-14
Fix the multiple builds

TsvetanMilanov:
When we have multiple devices connected and a device related command is executed we should build the project only once instead of building for each device.


Mitko-Kerezov:
LGTM :+1:

TsvetanMilanov:
run ci

PR #1559, opened 2016-09-14
Use WKWebView Engine for Cordova >=5.0.0

TsvetanMilanov:
Since there is an issue with our WKWebView for Cordova projects with version >=5.0.0 we need to use the WKWebView Engine from Cordova for those projects. When we set the webview we need to check the framework version and choose the correct WKWebView. Also when we migrate the project we should change the WKWebView.

Fixes: http://teampulse.telerik.com/view#item/321628


TsvetanMilanov:
run ci

TsvetanMilanov:
Closing this and re-target to release.

PR #1561, opened 2016-09-15
Keep the current indentation of json files

TsvetanMilanov:
Since our other clients use 2 spaces for indentation in json files we should use 2 spaces too. Since some users are using only the CLI we should not change their current indentation.

Merge after: https://github.com/telerik/mobile-cli-lib/pull/805


TsvetanMilanov:
run ci

Mitko-Kerezov:
Hoping for a green build here too 👍

PR #1563, opened 2016-09-15
Fix adding of NativeScript plugin

rosen-vladimirov:
NativeScript plugins have nativescript key in package.json based on which we decide if the plugin can be used with our current NativeScript version. Currently our check is incorrect, so any plugin which has nativescript key in it's package.json and the version matches with our NativeScript version cannot be added.


TsvetanMilanov:
:+1:

PR #1566, opened 2016-09-17
Validate hook arguments before execute

TsvetanMilanov:
Since we can have hooks with arguments which are invalid for the AppBuilder CLI (for example the before-livesync hook of nativescript-angular and watch hook of nativescript-dev-typescript) we need to check if the arguments of the hook are valid before we execute it. We can do the check only for hooks which can be executed in our process.

Merge after https://github.com/telerik/mobile-cli-lib/pull/806
Fixes: http://teampulse.telerik.com/view#item/321923


rosen-vladimirov:
:+1:

TsvetanMilanov:
run ci

tytodorov:
👍

TsvetanMilanov:
Closing this because the common lib sha will bi updated with https://github.com/Icenium/icenium-cli/pull/1570

PR #1567, opened 2016-09-17
Do not add fetched plugin to core plugins

TsvetanMilanov:
When we fetch plugin in Cordova project we must not add it to the core plugins because when we build the project the build will fail for duplicate symbols/classes.

Fixes http://teampulse.telerik.com/view#item/321807

### DIFF
--- a/lib/services/plugins/cordova-project-plugins-service.ts
+++ b/lib/services/plugins/cordova-project-plugins-service.ts
@@ -320,7 +320,13 @@ export class CordovaProjectPluginsService extends PluginsServiceBase implements 
 			pluginData.addPluginToConfigFile = false;
 
 			let pluginBasicInfo = super.installLocalPlugin(pathToInstalledPlugin, pluginData, options).wait();
-			this.configurePlugin(pluginBasicInfo.name, version).wait();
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
:+1:

ndinev:
:+1: Batbloking

lib/services/plugins/cordova-project-plugins-service.ts
```diff
@@ -415,22 +421,8 @@ export class CordovaProjectPluginsService extends PluginsServiceBase implements
...
-				}
-
-				_.each(variables, (variable: any) => {
-					let variableName: string = variable.name || variable;
-					let variableInformation = this.gatherVariableInformation(pluginData, variableName, configuration).wait();
-					cordovaPluginVariables[pluginData.Identifier][variableName] = variableInformation[variableName];
-				});
-				this.$project.setProperty(CordovaProjectPluginsService.CORDOVA_PLUGIN_VARIABLES_PROPERTY_NAME, cordovaPluginVariables, configuration);
-			}
+			this.setPluginVariables(pluginData.Identifier, <string[]>pluginData.Variables, configuration).wait();
```
Mitko-Kerezov: `configuration` can be undefined here, causing unexpected behavior

PR #1570, opened 2016-09-21
Transpile TS files on file change in livesync

TsvetanMilanov:
When we run the livesync command with the --watch flag the .ts files are not transpiled and there are no new .js files to sync to the device. We should transpile the .ts files before we get the files to sync and remove the .ts files from the sync queue to avoid transpiling twice.

Merge after: https://github.com/telerik/mobile-cli-lib/pull/808
Fixes: http://teampulse.telerik.com/view#item/322155


Mitko-Kerezov:
👍

TsvetanMilanov:
run ci

tytodorov:
👍

PR #1571, opened 2016-09-21
Respect npm settings - proxy and registry 

rosen-vladimirov:
Use npm proxy for connections to registry.npmjs.org. This way in case the user has set a proxy, we'll use it.
Also respect user's settings for npm registry.

http://teampulse.telerik.com/view#item/322319


ndinev:
:+1: This help to mitigate NPM connectivity issues

PR #1574, opened 2016-09-26
Fix check for ios simulator

TsvetanMilanov:
When we check if there is running simulator we use ps cax but if the simulator is killed for a short period of time after that there is a process listed when using ps cax and we decide that there is a running simulator and then we try to use it but ios-sim-portable returns null. We should check if the simulator returned from ios-sim-portable is not null or undefined in order to continue to work with it.


TsvetanMilanov:
run ci

TsvetanMilanov:
run ci

TsvetanMilanov:
Closing this because the common lib sha will be updated with https://github.com/Icenium/icenium-cli/pull/1579

PR #1576, opened 2016-09-27
Use WKWebView Engine for Cordova >=5.0.0

TsvetanMilanov:
Since there is an issue with our WKWebView for Cordova projects with version >=5.0.0 we need to use the WKWebView Engine from Cordova for those projects. When we set the webview we need to check the framework version and choose the correct WKWebView. Also when we migrate the project we should change the WKWebView.

Fixes: http://teampulse.telerik.com/view#item/321628

### DIFF
--- a/lib/services/cordova-migration-service.ts
+++ b/lib/services/cordova-migration-service.ts
@@ -226,6 +226,9 @@ export class CordovaMigrationService implements ICordovaMigrationService {
 			this.$logger.info("Migrating to Cordova version %s", versionDisplayName);
 			let oldVersion = this.$project.projectData.FrameworkVersion;
 			let availablePlugins = this.$pluginsService.getAvailablePlugins();
+
+			this.migrateWebView(oldVersion, newVersion).wait();
+
 			this.invalidMarketplacePlugins = _(this.$project.configurations)
 				.map(configuration => <string[]>this.$project.getProperty("CorePlugins", configuration))
 				.union()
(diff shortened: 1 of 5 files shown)


TsvetanMilanov:
run ci

lib/services/cordova-migration-service.ts
```diff
@@ -352,5 +355,19 @@ export class CordovaMigrationService implements ICordovaMigrationService {
 			return plugin;
 		});
 	}
+
+	private migrateWebView(oldFrameworkVersion: string, newFrameworkVersion: string): IFuture<void> {
+		return (() => {
+			// For Cordova versions below 5.0.0 with WKWebView we need to set the WKWebView to com.telerik.plugins.wkwebview.
+			// For Cordova version 5.0.0 and above with WKWebView we need to set the WKWebView to cordova-plugin-wkwebview-engine.
+			let currentWebViewName = this.$webViewService.getCurrentWebViewName(this.$projectConstants.IOS_PLATFORM_NAME);
+			let currentWebView = this.$webViewService.getWebView(this.$projectConstants.IOS_PLATFORM_NAME, currentWebViewName, oldFrameworkVersion);
```
rosen-vladimirov: You'll execute a lot of operations even in case the currentWebView is the same as the new one (for example migration from 3.7.0 to 4.0.0). Can we optimize this and do nothing in such cases?

tytodorov:
👍

PR #1577, opened 2016-09-27
Fix Ionic project transformation

TsvetanMilanov:
In the Ionic project template in the plugins directory is added fetch.json file. Before we get the information of the local plugins we need to check if the entry in the plugins directory is file or directory.

Fixes: http://teampulse.telerik.com/view#item/322371


TsvetanMilanov:
run ci

tytodorov:
:+1

PR #1583, opened 2016-10-03
Fix is ionic project check

TsvetanMilanov:
The ionic project configuration file is renamed from ionic.project to ionic.config.json. When we determine the type of the project in the check for ionic project we must check for both the old and the new file name.


tytodorov:
👍

PR #1584, opened 2016-10-04
Do not try to resolve hookArgs

TsvetanMilanov:
When we validate the hook arguments we try to resolve all of them but there is one argument which is not registered in the injector - hookArgs. If some hook wants to use it we will try to resolve it and fail. The solution is to skip the validation for the hookArgs argument.


rosen-vladimirov:
👍

TsvetanMilanov:
run ci

PR #1586, opened 2016-10-05
Change the submodule url to https

TsvetanMilanov:
We should use https instead of ssh for the mobile-cli-lib.


Mitko-Kerezov:
LGTM :+1:

TsvetanMilanov:
run ci

PR #1588, opened 2016-10-13
Fix remote command

TsvetanMilanov:
Add client to simulator name mapping for Xcode 8.

Also we must get appId from the request for users with Xcode 7+ in order to run the application on the iOS simulator.


TsvetanMilanov:
run ci

PR #1590, opened 2016-10-24
Fix Cordova plugins search

TsvetanMilanov:
Since there is some change in the npmjs search our current way of searching for plugins is broken. The problem is that now the order of the keywords in the search does matter.
For example if we search for "`ecosystem:cordova console`" we won't find any plugins but if we search for "`console ecosystem:cordova`" we will find the plugins that we are looking for.


TsvetanMilanov:
Closing this because it must be re-targeted to release.

PR #1591, opened 2016-10-26
Add support for Node 7

rosen-vladimirov:
Add support for Node 7:
 - update dependencies to C++ modules
 - update engine requirements in package.json
 - update temp and istanbul to prevent deprecation message for os.tmpDir


TsvetanMilanov:
👍

PR #1598, opened 2016-11-10
Fix cannot find module ios-log-filter

TsvetanMilanov:
Change the path to ios-log-filter to point to the correct folder.

Merge after https://github.com/telerik/mobile-cli-lib/pull/837


TsvetanMilanov:
Closing this and re-target it to release.

PR #1608, opened 2016-11-28
Transpile to ES6

rosen-vladimirov:
Add support for ES6 transpilation. Change dependency injection in order to match the `class` and `constructor` keywords.
Update istanbul to latest version as its previous versions do not support ES6.

Remove bluebird.d.ts as it conflicts with Promises from TypeScript's ES6 d.ts.
Remove bluebird as dependency - we will use native Promises from now on.

Add support for Node.js 4 and 5 - both versions have limited support for ES6 features, so in order to support them we have to:
 - pass --harmony flag to Node.js
 - do not use some features like default values of method arguments.
 - do not use spread operator - it's implementation in Node.js 4 is limited only to arrays. Even this implementation has issues. So use `.apply` instead of using spread operator. Remove all other usages of spread operator (for objects).

Add tests scripts for istanbul and mocha, as they both will need the `--harmony` flag as well, so we cannot execute them directly with Node.js 4 and 5. Instead we'll execute a custom script, that will start a new node process.

Introduce `test-scripts` directory with JavaScript executables to start the tests. Set them in package.json. They are required as when Node 4 or Node 5 is used, the tests will need `--harmony` flag as well.
Exclude `test-scripts` directory from npm package (add it to `.npmignore`) - users do not need these files.

Exclude `.vscode/launch.json` from `.gitignore` and add predefined launch configurations. They can be used for debugging CLI process and unit tests.

When a package has to be used globally, npm automatically creates required executable based on the `bin` key in `package.json`.
In our case the executable, that's created, is a shell script (`.cmd` on Windows) that starts a node process and launches `<cli dir>/bin/appbuilder.js`.
This makes it impossible for us to modify the arguments passed to Node.js. But for Node.js 4.x.x, we want to pass `--harmony` flag.
In order to fix this, modify appbuilder.js to checks current node version and if we want to pass additional arguments to Node.js process, start a new Node process inside it.
In case the version of Node is 6.x.x or later, we'll use the old code, where the operations are in the same process.

For easier development, we've created several shell scripts (`.cmd` on Windows) inside CLI's bin directory. Developers of CLI add the bin dir to their `$PATH` variable instead of installing the CLI globally. In order to allow usage of `tns` directly from the terminal for them, we've created shell scripts that start the `bin/appbuilder.js` executable.
However the `bin/appbuilder.js` has shebang header and it can be executed directly. So in order to simplify the bin directory, rename `bin/appbuilder.js` to `tns` and remove all other files.
This way when CLI developers add bin to their $PATH environment variable, they'll be able to call `tns` directly. Set the executable name in `package.json`, so when the CLI is installed globally, everything will work as before.


TsvetanMilanov:
run ci

rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #1612, opened 2016-11-30
Use TypeScript 2.0

TsvetanMilanov:
Update the TS related dependencies to use TS 2.0 and fix the errors from
the update

Common lib reference: https://github.com/telerik/mobile-cli-lib/pull/846

TP reference: http://teampulse.telerik.com/view#item/325921


package.json
```diff
@@ -74,7 +74,7 @@
     "shelljs": "0.5.3",
     "tabtab": "https://github.com/Icenium/node-tabtab/tarball/master",
     "temp": "0.8.3",
-    "typescript": "1.8.10",
+    "typescript": "2.0.7",
```
rosen-vladimirov: why not 2.0.10 ?

AssiNET:
👍

PR #1613, opened 2016-12-01
Fix iOS 10 device logs

TsvetanMilanov:
On iOS 10 devices the device logs contain \n after each line.
When we use readCString for buffers which contain many \0
the method will return the content of the buffer only before the first \0 character.
We need to replace the \0 with "" in order read the whole content.

TP item: http://teampulse.telerik.com/view#item/325955
Common lib reference: https://github.com/telerik/mobile-cli-lib/pull/848 (Old PR)
Common lib reference: https://github.com/telerik/mobile-cli-lib/pull/850 (New PR)


ndinev:
:+1:

TsvetanMilanov:
run ci

PR #1615, opened 2016-12-05
Break CLI process when Node is not supported

rosen-vladimirov:
As old versions of Node.js cannot be used anymore (after transpiling to ES6), we should break the process when such case happens.
Use only var instead of let, const, as only var keyword is supported in Node.js 0.10.x


rosen-vladimirov:
run ci

rosen-vladimirov:
run ci

PR #1620, opened 2016-12-13
Set stdio inherit to npm install

TsvetanMilanov:
We need to set the stdio when installing node modules because nativescript-telerik-ui asks for email in its post-install script and does not check if the terminal is interactive or not. If it is not interactive it will ask for email and out npm install will hang.

Common lib reference: https://github.com/telerik/mobile-cli-lib/pull/854.
TP reference: http://teampulse.telerik.com/view#item/326658.


ndinev:
:+1: This problem results in blocking problem - user can't create NS Angular empty project in CLI

PR #1635, opened 2017-02-06
Do not open AppManager URL in browser when --skipUi is passed

rosen-vladimirov:
For automation purposes we should not open AppManager URL in the browser.
This will allow users to automate commands like `$ appbuilder appmanager upload <platform>`
and `$ appbuilder appmanager livesync <platform>`
So in case `--skipUi` is passed to these commands, all actions will be executed, but the last step (opening the URL in browser) will not be executed.


rosen-vladimirov:
run ci

PR #1636, opened 2017-02-06
Remove ios native dependencies

TsvetanMilanov:
We need to remove our native dependencies - ref and ffi because they are casuing problems with each version of NodeJS which has new version of V8 in it. We are using ref and ffi to communicate with iTunes and execute fs and application management operations on iOS devices. This operations are moved in new dependency - device-lib. The device lib has JavaScript wrapper which we use in the CLI. Everything in the ios device file system and ios device application manager is replaced with the exposed methods from the device lib.

Common lib reference: https://github.com/telerik/mobile-cli-lib/pull/889


TsvetanMilanov:
run ci

PR #1639, opened 2017-02-16
Fix check if package has `@types`

rosen-vladimirov:
During migration from fibers to promises, we've lost one `!` in the check if npm package has typings in `@types` repo.
This led to incorrect behavior:
 - typings of packages, which are available in `@types`, are not installed (for example `lodash`).
 - trying to install plugin, which does not have typing in `@types` leads to error (something like `Error while installing dependency: Command npm.cmd failed with exit code 4294963238.`).

Fix the check (add the missing `!`) which will fix the incorrect behavior.


TsvetanMilanov:
run ci

TsvetanMilanov:
run ci

PR #1640, opened 2017-02-21
Fix building of the package 

rosen-vladimirov:
Currently building the package fails when CLI tries to download `https://<environment>/appbuilder/Resources/NativeScript/tns_modules/TS/2.5.0tns-core-modules.zip`
As we are now using tns-core-modules 2.5.1 the zip name that CLI tries to download is incorrect (should be 2.5.1). This leads to error, so other resources AFTER this one are not downloaded.
So the produced package is not correct.
As these zip files had been used in older CLI versions, skip their download. They have been required when we've been using local `.d.ts` files for `tns-core-modules`. Now we rely on the `.d.ts` files in the npm package, so no need ot include any of the `.zip` files in CLI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,10 @@ The AppBuilder CLI lets you build, test, deploy, and publish cross-platform hybr
 Installation
 ===
 
-Latest version: AppBuilder 3.6.7
-<br/>Release date: February 2, 2016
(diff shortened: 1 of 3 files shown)


ndinev:
👍 go go

PR #1643, opened 2017-02-22
Fix transpilation problems

TsvetanMilanov:
The first problem is when in the tsconfig.json file in the compilerOptions there is a property with array value. We need to pass the array values correctly to the tsc command.

The second problem is with TypeScript 2.1 output. When there is an error in the output while transpiling TS 2.1 will make the `error` word red. This means that the actual autput will look like this: `[91merror [0m TS2304`. We need to update our regular expressions to match the new output.

Common lib reference: https://github.com/telerik/mobile-cli-lib/pull/897
Fixes: http://teampulse.telerik.com/view#item/329315


ndinev:
👍

TsvetanMilanov:
run ci

PR #1644, opened 2017-02-22
Fix prompting for plugin variables

rosen-vladimirov:
When a plugin has variables that has to be set, CLI checks the command line and in case the `--var` option is not passed, a prompt is shown to the user to enter a value.
At the moment this is broken as we are prompting in parallel promises. This leads to several issues - user is not able to enter values for the variables and also a memory leak warning is shown.

Fix the execution to be sync.

Bug is reproducible for both Cordova and NativeScript projects.


ndinev:
👍

PR #1649, opened 2017-03-16
Fix device log filter

TsvetanMilanov:
We have added one parameter in the filter method which we do not need and we do not respect it when we call the method. The solution is to remove it.

Common lib reference: https://github.com/telerik/mobile-cli-lib/pull/905


rosen-vladimirov:
run ci

PR #1651, opened 2017-03-23
Fix expected Cordova version in tests to 6.4.0

rosen-vladimirov:
Also update `grunt-shell` as the old version fails to run the tests with `maxbuffer exceeded`.


rosen-vladimirov:
run ci

PR #1658, opened 2017-03-31
Use request module for http requests

Mitko-Kerezov:
Merge after https://github.com/telerik/mobile-cli-lib/pull/928


Mitko-Kerezov:
TP Ref: [[CLI] Provide a way to set credentials for proxy](http://teampulse.telerik.com/view#item/330417)

docs/man_pages/general/proxy-set.md
```diff
@@ -3,13 +3,15 @@ proxy set
 
 Usage | Synopsis
 ------|-------
-General | `$ appbuilder proxy set [<Hostname> [<Port><% if(isWindows) {%> [<Username> [<Password>]]<%}%>]]`
+General | `$ appbuilder proxy set [<Url> <% if(isWindows) {%>[<Username> [<Password>]]<%}%>]`
 
 Sets proxy settings.
 
+### Options
+* `--insecure` - Allows insecure SSL connections and transfers to be performed. All SSL connections are attempted to be made secure by using the CA certificate bundles installed. All connections considered "insecure" will fail unless this flag is used.
```
rosen-vladimirov: This seems like I have to set this flag in cases when I want to access insecure sites, however CLI decides what to access. We have to explain that the flag has to be used when the proxy itself does not have valid SSL certificate

PR #1664, opened 2017-04-11
Set release date

rosen-vladimirov:
Set release date.
Add information for users that are aware of the old way to set proxy. When they try to set the USE_PROXY value to true, they'll find how to set the proxy.


config/config-base.json
```diff
@@ -2,10 +2,8 @@
 	"AB_SERVER_PROTO": "http",
 	"AB_SERVER": "localhost",
 	"DEBUG": true,
-	"USE_PROXY": false,
+	"USE_PROXY": "Not supported anymore. For more information check our documentation: http://docs.telerik.com/platform/appbuilder/development-tools/running-appbuilder/running-the-cli/downloading-and-installing-cli",
```
tytodorov: Not descriptive enough.

PR #1671, opened 2017-04-21
Cloud export missing await

Mitko-Kerezov:
Add missing await - else `cloud export` fails because we try to unzip before downloading

Ping @rosen-vladimirov 
TP: [[CLI] Cloud export command fails](http://teampulse.telerik.com/view#item/344205)


nadyaA:
run ci

PR #1682, opened 2017-08-10
Warn instead of fail when QR code is too long

Mitko-Kerezov:
Print a warning instead of failing when QR code is too long.


Mitko-Kerezov:
Discussed this offline and decided the PR is no longer relevant, hence I'm closing it

PR #1691, opened 2017-09-27
Update ios-device-lib to latest version

rosen-vladimirov:
Update `ios-device-lib` to latest version, where several bugfixes are included.


ndinev:
👍

PR #1695, opened 2017-11-06
Introduce export command

Mitko-Kerezov:
Introduce `$ appbuilder export` command, which enables exporting a project for development with Cordova/NativeScript CLI.

Ping @rosen-vladimirov @nadyaA @TomaNikolov

### DIFF
--- a/lib/commands/cloud-projects.ts
+++ b/lib/commands/cloud-projects.ts
@@ -57,7 +57,7 @@ export class CloudListProjectsCommand implements ICommand {
 $injector.registerCommand("cloud|*list", CloudListProjectsCommand);
 
 export class CloudExportProjectsCommand implements ICommand {
-	protected _shouldCheckProject = true;
+	protected _isCloudToLocalExport = true;
 	constructor(protected $errors: IErrors,
 		protected $remoteProjectService: IRemoteProjectService,
 		protected $cloudProjectsService: ICloudProjectsService,
@@ -85,13 +85,17 @@ export class CloudExportProjectsCommand implements ICommand {
 			this.$errors.failWithoutHelp("You do not have any projects in the cloud.");
 		}
 
-		if (this._shouldCheckProject && this.$project.projectData) {
+		if (this._isCloudToLocalExport && this.$project.projectData) {
 			this.$errors.failWithoutHelp("Cannot create project in this location because the specified directory is part of an existing project. Switch to or specify another location and try again.");
 		}
 
 		if (args && args.length) {
 			if (args.length > 2) {
 				this.$errors.fail("This command accepts maximum two parameters - solution name and project name.");
+			}
+
+			if (!this._isCloudToLocalExport && !commonHelpers.isInteractive() && (!args[0] || !args[1])) {
+				this.$errors.fail("Running this command in non-interactive mode requires both parameters - solution name and project name.");
 			}
 
 			let slnName = args[0];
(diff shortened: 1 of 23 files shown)


lib/commands/cloud-projects.ts
```diff
@@ -117,5 +111,73 @@ export class CloudExportProjectsCommand implements ICommand {
...
+					this.$fs.ensureDirectoryExists(dirname(downloadPath));
+				} else {
+					this.$fs.ensureDirectoryExists(downloadPath);
+					downloadPath = join(downloadPath, buildItem.Filename);
+				}
+			}
+
+			this.$logger.info(`Downloading ${buildItem.Filename} to ${downloadPath}...`);
+			const targetFile = this.$fs.createWriteStream(downloadPath);
+			await this.$httpClient.httpRequest({
```
rosen-vladimirov: In case this request fails, it will still create a targetFile with size of 0KB. I suggest to try - catch this call and delete the targetFile in case operation fails.

lib/declarations.d.ts
```diff
@@ -164,7 +165,7 @@ interface ILoginManager {
 	telerikLogin(user: string, password: string): Promise<void>;
 }
 
-declare module Server.Contract {
+declare module Server2.Contract {
```
rosen-vladimirov: why?

lib/services/cloud-projects-service.ts
```diff
@@ -0,0 +1,53 @@
...
+
+	public async getSolutionProjectInfo(opts: ISolutionProjectPairOptions): Promise<ISolutionProjectInfo> {
+		const apps = await this.$remoteProjectService.getAvailableAppsAndSolutions();
+		let framework = "";
+		let id = "";
+		const noSolutionNamePassed = !opts.solutionName;
+
+		if (noSolutionNamePassed) {
+			const solutionNames = apps.map(sln => sln.colorizedDisplayName) || [];
+			opts.solutionName = await this.$prompter.promptForChoice("Select solution to export", solutionNames);
```
rosen-vladimirov: what if the terminal is not interactive?

lib/services/cloud-projects-service.ts
```diff
@@ -0,0 +1,53 @@
...
+			this.$errors.failWithoutHelp(`Solution ${solutionData.Name} does not contain any projects.`);
+		}
+
+		if ((noSolutionNamePassed && !opts.projectName) || (opts.forceChooseProject && !opts.projectName)) {
+			const projects = (await this.$remoteProjectService.getProjectsForSolution(opts.solutionName)).map(proj => proj.Name);
+			if (opts.enableExportWholeSolution) {
+				projects.push(ExportOptions.WholeSolution);
+			}
+
+			const selection = await this.$prompter.promptForChoice("Select project to export", projects);
```
rosen-vladimirov: same question for non-interactive terminal

lib/declarations.d.ts
Mitko-Kerezov: honest mistake

lib/services/cloud-projects-service.ts
Mitko-Kerezov: this is handled in the command's `canExecute` method

lib/services/cloud-projects-service.ts
Mitko-Kerezov: this could be a problem - will fix

Mitko-Kerezov: fixed with the latest commit

